PR #43, opened 2025-04-14
Add new Board an Config

k2dp2k:


Sleeper85:
Hi,

Is this a PCB that you printed or something that is available commercially ?

image

k2dp2k:
Hi thats one i printed, espacially for Deye and JK PB Series

Sleeper85:
Regarding the board name, you should find a different name.
It's just a board with an RS485 and CAN port, so it can be used with other devices besides Deye and JK.

If it's an ESP32-S3, the name should be `board_ESP32-S3_BoardName.yaml`.

Sleeper85:
Are the other GPIOs available ?

Or is there only an RS485 and CAN port ?

PR #44, opened 2025-04-16
Add support for LilyGo T-Connect Pro

GHswitt:
This adds support for the LilyGo T-Connect Pro

I created this PR early so others can see that it is work in progress.

https://lilygo.cc/products/t-connect-pro 
https://github.com/Xinyuan-LilyGO/T-Connect-Pro/tree/main

TODO: 

- ~Display pages~ Disabled by default as updating it increases the loop time too much
- ~Relais~
- ~Alternative configuration file: Ethernet instead of display (ESPHome does not support ethernet and other SPI devices on the same SPI bus)~

I got it up and running as Modbus client:
IMG_20250416_182043
Modbus server is a LilyGo T-Connect with Deye (CAN) and JK (RS485) connected (more to come):
IMG_20250416_182104

### DIFF
--- a/packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml
+++ b/packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml
@@ -1,0 +1,150 @@
+# Updated : 2025.04.16
+# Version : 1.1.1
+# GitHub  : https://github.com/Sleeper85/esphome-yambms
+
+# YamBMS ( Yet another multi-BMS Merging Solution )
+
+# This YAML is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation, either version 3
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/gpl.html>.
+
+# https://github.com/Xinyuan-LilyGO/T-Connect-Pro
+
+substitutions:
+  board_chip: "ESP32-S3"
+  board_name: "LilyGo T-Connect Pro"
+  # uart_esp_1: RS232
+  tx_pin_1: '4'
+  rx_pin_1: '5'
+  # uart_esp_2: RS485
+  tx_pin_2: '17'
+  rx_pin_2: '18'
+  # canbus_esp32_can: CAN
+  tx_pin_4: '6'
+  rx_pin_4: '7'
+
+packages:
+  device_base: !include ../base/device_base.yaml
+  device_base_wifi: !include ../base/device_base_wifi.yaml
+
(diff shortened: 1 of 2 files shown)


GHswitt:
@Sleeper85 I've a problem with the internal CAN controller & display enabled. With display enabled:

`  display: !include board_options_display_480x222.yaml`

I get in the logs:

`Component canbus is marked FAILED`

(Which seems to be marked failed during TWAI driver installing & starting)
And the CAN bus to the inverter is not working. If I remove the display package it works. Any idea? Too slow with display?

Sleeper85:
> @Sleeper85 I've a problem with the internal CAN controller & display enabled. With display enabled:
> 
> ` display: !include board_options_display_480x222.yaml`
> 
> I get in the logs:
> 
> `Component canbus is marked FAILED`
> 
> (Which seems to be marked failed during TWAI driver installing & starting) And the CAN bus to the inverter is not working. If I remove the display package it works. Any idea? Too slow or not enough memory with display?
> 
> Debug info, without display:
> 
>     * Inverter heartbeat is ~600ms with Pylon 1.2 and ~1400ms with Pylon V2
> 
>     * Heap free 48%
> 
>     * Loop time 20-70ms
> 
> 
> With display:
> 
>     * Heap free 14%
> 
>     * Loop time ~130ms

No real idea, sorry.
Maybe you need to find better options for PSRAM and not import the YAML `board_options_psram_settings.yaml`.
Look at what I did for the `Waveshare Touch LCD screen` in the YAML `board_ESP32-S3_Touch-LCD-7.yaml`.

Simply enabling PSRAM consumes DRAM resources, which is why I disable it by default because unfortunately it doesn't improve performance and is even worse.

A test I performed on an ESP32-S3.
I can monitor 3 JK-BMS via BLE without enabling PSRAM and only 2 with PSRAM enabled.

GHswitt:
I think because of the display resolution, even with lower color_palette, it just needs much memory. With PSRAM it works, but updating the display is slow and will take the loop time above 100ms. So I leave the display disabled by default, but others are free to enable and use it.

Other than that the board is working fine now for several weeks.
I've updated the PR to the latest main.

Sleeper85:
I've simplified the `board.yaml` file for use with remote packages.

PSRAM activation will be done in `main.yaml` by importing `packages/board/board_options_ESP32-S3_PSRAM.yaml`.

For `T-Connect-Pro`, the easiest way would be to import `board_options_ESP32-S3_PSRAM.yaml` into your YAML `board_options_display_480x222.yaml`.

The board can be imported in `main.yaml` like this : 

```YAML
  board:
    ref: main
    refresh: 0s
    files:
      - path: 'packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml'
        vars:
          display_auto_next_page_interval: '5s'
          display_rotation: '0' # 0/90/180/270 rotation angle in degrees
      # Importing the display also activates PSRAM.
      # Also note that the display update takes about 120 ms, as indicated in the logs:
      # "Component display took a long time for an operation (124 ms)."
      - path: 'packages/board/board_options_display_480x222.yaml'
```

Please, edit `board_ESP32-S3_LilyGo-T-Connect-Pro.yaml` and remove the lines :
* 38 to 43
* 57 to 61
* 67
* 85

Thanks

GHswitt:
Thank you for the clarification, I've update the files accordingly.

PR #45, opened 2025-04-22
Add Modbus server files for JK BMS (RS485) and DEYE (CAN)

GHswitt:
This adds two Modbus server files, one for JK-BMS connected via RS485 and one for DEYE connected via CAN.
With these files the mentioned BMS can be used as Modbus node.


Sleeper85:
Please name the YAML files like this:

- bms_modbus_DEYE_CAN_module_full.yaml
- bms_modbus_JK_RS485_Modbus_bms_standard.yaml

GHswitt:
Done, thank you.

PR #50, opened 2025-05-15
DEYE: Add throttle to some values

GHswitt:
Throttle Deye operating mode and requested charging/discharging current as it fluctuates a lot and therefore causes lots of updates.


Sleeper85:
Can you adapt the date and version at the top of the YAML ?

Thanks ;-)

GHswitt:
Done thanks :)

PR #53, opened 2025-05-22
Add feature "Auto float voltage"

GHswitt:
## Auto Float Voltage

When switching from `Bulk` to `Float`, therefore lowering the `Requested Charge Voltage (CVL)`, some inverters (e.g. Deye SUN-12K) will start to discharge the battery (into the grid) with up to max. discharge current until the
`Requested Charge Voltage (CVL)` is reached. The reason behind this is that the inverter sees this as battery overvoltage and wants to counter it. 
As workaround, by enabling `Automatic Float Voltage`, it will lower the `Requested Charge Voltage (CVL)` gradually in small steps until the float voltage is reached, allowing the battery more time to settle down.
There are two modes available: `Follow` and `Immediately`.

- `Follow`: Waits until the battery voltage itself is below the current `Requested Charge Voltage (CVL)` before lowering `Requested Charge Voltage (CVL)` further. It follows the battery voltage until the target float voltage is reached.
- `Immediately`: Immediately start lowering the `Requested Charge Voltage (CVL)`, independent of the current battery voltage. That means it can go below the current battery voltage.

This feature might not work with all inverters as good or is not required if these do not discharge the battery.

The voltage steps can be configured (default `0.1V`) with `Auto Float Voltage Step`. The update interval can be configured (default `3 minutes`) with `Auto Float Update Interval`.

With a 2 minute time interval, `Follow` mode looks like this:

image

And with one minute time interval. It will still wait for the voltage to drop but it will change the `RCV` faster.
image

Note: I've a `Inverter Offset V.` of 0.2V.

### DIFF
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
@@ -247,6 +247,16 @@ views:
         title: YamBMS - Auto CCL / DCL
       - type: entities
         entities:
+          - entity: select.yambms_demo_yambms_1_automatic_float_voltage
+            name: Automatic Float Voltage
+          - entity: sensor.yambms_demo_yambms_1_auto_float_voltage
+            name: Auto Float Voltage
+          - entity: number.yambms_demo_yambms_1_auto_float_update_interval
+          - entity: number.yambms_demo_yambms_1_auto_float_voltage_step
+            name: Auto Float Voltage Step
+        title: YamBMS - Auto Float
+      - type: entities
+        entities:
           - entity: number.yambms_demo_yambms_1_rebulk_soc
             name: Rebulk SoC
           - entity: number.yambms_demo_yambms_1_rebulk_v
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_JK_RS485_3xBMS.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_JK_RS485_3xBMS.yaml
@@ -247,6 +247,16 @@ views:
         title: YamBMS - Auto CCL / DCL
       - type: entities
         entities:
+          - entity: select.yambms_yambms_1_automatic_float_voltage
+            name: Automatic Float Voltage
+          - entity: sensor.yambms_yambms_1_auto_float_voltage
+            name: Auto Float Voltage
+          - entity: number.yambms_yambms_1_auto_float_update_interval
+          - entity: number.yambms_yambms_1_auto_float_voltage_step
+            name: Auto Float Voltage Step
+        title: YamBMS - Auto Float
+      - type: entities
+        entities:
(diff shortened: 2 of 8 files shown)


Sleeper85:
I had already read your code, I probably have questions to ask you, I will check again later. Thank you

Sleeper85:
The `yambms.yaml` file you modified isn't the latest version.

I will upgrade the YamBMS version to `1.5.7` when I release this new feature.

I tested your `Auto Float` function and reorganized your code so that the expected behavior is clearer to me.

```C++
      // Variables
      const auto mode = id(${yambms_id}_auto_float_mode).active_index();        // Mode: 0 - Disabled, 1 - Follow, 2 - Immediately
      const double bulk_v = id(${yambms_id}_bulk_voltage).state;                // Bulk voltage
      const double target_float_v = id(${yambms_id}_float_voltage).state;       // Target float voltage
      const double step_v = id(${yambms_id}_auto_float_step_v).state;           // Voltage step for each decrement
      const double max_total_v = id(auto_float_total_voltage).state;            // Total voltage (mean and max)
      const auto update_interval = id(${yambms_id}_auto_float_interval).state;  // Update interval in minutes
      static double float_v = bulk_v;                                           // Initialize float voltage
      static uint8_t time_count = 0;                                            // Interval counter
      bool decrement = false;
      
      // Auto Float disabled OR Bulk voltage <= Float voltage
      if ((mode == 0) || (bulk_v <= target_float_v))
      {
        // Return Float voltage
        id(${yambms_id}_auto_float) = 0;
        return target_float_v;
      }
      // Auto Float enabled + Floating
      else if (id(${yambms_id}_charge_status) == "Float")
      {
        // Check update interval
        if (++time_count >= update_interval)
        {
          // Follow mode : Wait for battery to drop below current float voltage
          if (mode == 1)
          {
            // Check battery voltage and wait for the battery voltage to reach current float voltage
            if ((id(${yambms_id}_total_voltage).state > 0) && (max_total_v <= float_v)) decrement = true;
          }
          // Immediately mode : Start immediately
          else if (mode == 2) decrement = true;

          // Decrement Float voltage
          if (decrement && (float_v > target_float_v)) float_v -= step_v;
          // Stop Decrement
          if (float_v < target_float_v) float_v = target_float_v;

          // Reset update interval counter
          time_count = 0;
        }
      }
      // Auto Float enabled + No Floating
      else
      {
        // Reset values
        id(${yambms_id}_auto_float) = 0;
        float_v = bulk_v;
        time_count = 0;
      }

      // Store difference for CVL calculation
      id(${yambms_id}_auto_float) = float_v - target_float_v;

      // Return Auto Float voltage
      return float_v;
```

Sleeper85:
I suppose it's [you](https://diysolarforum.com/members/watthamster.152511/) ?

Sleeper85:
Released in YamBMS `1.5.7`.

GHswitt:
Thanks, looks good, I'll try it in the next days. Yes, that is me :)

PR #54, opened 2025-05-24
Update yambms_service.yaml

GeorgSixt:
headers has been replaced by request_headers


Sleeper85:
Yes, I'm aware of this, and the YAML has already been corrected for me, but this only works with esphome 2025.5.0 and later, and I didn't want to force users to update.

In addition, there are other points to check and components to correct for proper operation with esphome 2025.5.0.

For now, I recommend sticking with esphome 2025.4.2.

PR #60, opened 2025-06-10
Fixes for new ESPHome and add of BMS Cycle Count offset #2

k2dp2k:


k2dp2k:
sorry stupid me....  i have overwritten the version numbers in all the files... -.-

k2dp2k (commented):
That one is new in this version

PR #67, opened 2025-06-29
Add options to restrict max. charge and discharge current

GHswitt:
With multiple battery packs, my max. charge current was somewhere about 700A+. However my inverter only supports about 240A. While this did not create function problems, it is hard to control the charge and discharge current with
'Max charge current (%)' and 'Max discharge current (%)' because the usable range was about 1-20%.

This adds 'yambms_max_inverter_charge_current' and 'yambms_max_inverter_discharge_current' substitutions. They are used to restrict the range of the charge/discharge current for easier control when the battery currents are much higher than the inverter currents. With this the full range of the mentioned control sliders/number can be better used.

### DIFF
--- a/YamBMS_LP_BASEN_RS485.yaml
+++ b/YamBMS_LP_BASEN_RS485.yaml
@@ -1,5 +1,5 @@
-# Updated : 2025.06.26
-# Version : 1.5.7
+# Updated : 2025.06.29
+# Version : 1.5.8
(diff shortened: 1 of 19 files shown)


GHswitt:
Done, thank you. I've updated all files that were modified.

GHswitt:
I rebased it onto the current main.

PR #74, opened 2025-07-26
RS485 Heartbeat

Fahmula:
This PR adds inverter heartbeat monitoring support for RS-485, closely modeled after the existing CANbus 0x305 implementation.

It introduces:
sensor.inverter_heartbeat — measures time between inverter responses (in ms)
binary_sensor.inverter_online — flags online/offline based on response timeout
switch.rs485_switch_inverter_heartbeat_monitoring — enables or disables tracking

Everything works well so far, but I wanted to raise a point for discussion:

The current Home Assistant dashboard examples assume hardcoded entity names like yambms_canbus_1_status, which means RS-485 users will need to manually update those to something like inverter_online or ${rs485_id}_status.

Would it make sense to:
Generalize the dashboards to use more flexible IDs (e.g. templated ${protocol_id}_status) so they can work for both CAN and RS-485?
Or should we keep the current dashboards CAN-specific, and just add a note or alternate example for RS-485 users?
In the future, do you see value in consolidating the CAN and RS-485 heartbeat logic into a shared proxy abstraction, or should they remain parallel?

### DIFF
--- a/packages/yambms/yambms_rs485_pylon.yaml
+++ b/packages/yambms/yambms_rs485_pylon.yaml
@@ -1,5 +1,5 @@
 # Updated : 2025.08.02
-# Version : 1.2.2
+# Version : 1.3.0
 # GitHub  : https://github.com/fahmula/esphome-pylontech-rs485
 
 # YamBMS ( Yet another multi-BMS Merging Solution )


Sleeper85:
The best thing would be to have a common code with a single dashboard. I'll think about it.

Sleeper85:
In reality, it's not necessary to create common sensors.

Let's not forget that the parameters of the users' protocol choices are saved and this change would require re-configuring the protocol after an upgrade.

We could even have a configuration with `CANBUS` + `RS485` at the same time.

We will keep your `PR` as proposed, you just need to make some adjustments (Inverter Online Status name).

Simply adjust the dashboard to display the available sensors with `entity-filter` based on the packages imported by the user.

Before

```YAML
      - type: entities
        entities:
          - entity: select.yambms_canbus_1_bms_name
            name: CANBUS 1 BMS Name
          - entity: select.yambms_canbus_1_protocol
            name: CANBUS 1 Protocol
        title: YamBMS - CANBUS Protocol
```

```YAML
      - type: entities
        entities:
          - entity: binary_sensor.yambms_canbus_1_status
            name: CANBUS 1 Status
          - entity: switch.yambms_canbus_1_inverter_heartbeat_monitoring
            name: CANBUS 1 Inverter Heartbeat Monitoring
          - entity: sensor.yambms_canbus_1_inverter_heartbeat
            name: CANBUS 1 Inverter Heartbeat
        title: YamBMS - Inverter Heartbeat
```

After

```YAML
      - type: entity-filter
        entities:
          - entity: select.yambms_canbus_1_bms_name
            name: CANBUS 1 BMS Name
          - entity: select.yambms_canbus_1_protocol
            name: CANBUS 1 Protocol
          - entity: select.yambms_rs485_1_protocol
            name: RS485 1 Protocol
        conditions:
          - condition: state
            state_not: unavailable
        card: 
          title: YamBMS - Inverter Comm. Protocol
```

```YAML
      - type: entity-filter
        entities:
          - entity: binary_sensor.yambms_canbus_1_status
            name: CANBUS 1 Status
          - entity: switch.yambms_canbus_1_inverter_heartbeat_monitoring
            name: CANBUS 1 Inverter Heartbeat Monitoring
          - entity: sensor.yambms_canbus_1_inverter_heartbeat
            name: CANBUS 1 Inverter Heartbeat
          - entity: binary_sensor.yambms_rs485_1_status
            name: RS485 1 Status
          - entity: switch.yambms_rs485_1_inverter_heartbeat_monitoring
            name: RS485 1 Inverter Heartbeat Monitoring
          - entity: sensor.yambms_rs485_1_inverter_heartbeat
            name: RS485 1 Inverter Heartbeat
        card: 
          title: YamBMS - Inverter Status & Heartbeat
```

Regarding the `online` status in any case, you must also update this `binary_sensor` based on the status of your `RS485 link`. This is used by boards like the `AtomS3` with a screen to display the status of the connection to the inverter.

```C++
id(inverter_com_status).publish_state(false);                   // Set Inverter com. status to OFF
```

The easiest way would be to also use this `binary_sensor` on the main dashboard page.
If you're using `CANBUS` and `RS485` simultaneously, you'll need to go to the `Control` page to view the two statuses separately.

Fahmula:
I just looked at the commits you have made and I will update my PR to be compatible. Great work!

Fahmula:
> Regarding the `online` status in any case, you must also update this `binary_sensor` based on the status of your `RS485 link`. This is used by boards like the `AtomS3` with a screen to display the status of the connection to the inverter.
> 
> ```c++
> id(inverter_com_status).publish_state(false);                   // Set Inverter com. status to OFF
> ```
> 
> The easiest way would be to also use this `binary_sensor` on the main dashboard page. If you're using `CANBUS` and `RS485` simultaneously, you'll need to go to the `Control` page to view the two statuses separately.

I see what you mean. With CAN it’s straightforward  we can just listen for can_id: 0x305 and directly update inverter_com_status when ACKs are received.

For RS485 it’s a bit trickier. I could simply set the inverter_com_status binary_sensor to true whenever the heartbeat switch is enabled, but that wouldn’t reflect the actual state of the RS‑485 link  only that monitoring is turned on. To represent the real status, we’d need to tie it to successful responses from the inverter (e.g. based on the heartbeat interval logic).

That would likely mean adjusting the RS‑485 component itself (not just the YAML) so we can reliably detect when the link is active and publish the state. Do you think that approach makes sense, or would you prefer I start with the simpler method (switch‑based) and leave deeper link‑level detection for a follow‑up?

Fahmula:
Just to follow up on my earlier comment:

In my RS485 component I’m already tracking is_data_valid_ and last_update_ms_ inside the loop, so I can publish a proper inverter_com_status binary_sensor without depending on the heartbeat switch.

The plan would be:

Set inverter_com_status to true once valid frames are received

Flip it back to false if no updates arrive within the configured update_timeout.

That way, the state reflects the actual RS485 link status instead of just whether monitoring is enabled. It should give dashboards the same behavior they currently get from CAN (with the 0x305 ACK).

If you agree, I can extend this PR to wire that into the component so dashboards can show CAN and RS485 consistently.

Sleeper85:
I completely agree.

The RS485 link status should not be related to the `Inverter Heartbeat`, which measures the pulses.

Everything I did with CAN bus stems from the fact that the CAN link had to be paused if the inverter didn't send an ACK, otherwise the ESP32 would end up crashing.

I created the `Inverter Heartbeat` function to check why the CAN bus link was disconnecting every two hours with my Deye, and this revealed that it was my Deye itself that was delaying its response every two hours, and even more so at midnight. Now I know that I need to adjust my link timer to 6 seconds with Deye.

This `Inverter Heartbeat` function is essentially a DEBUG function.

For RS485, it's the inverter itself that requests information from the BMS, so yes, you can monitor that => is the inverter sending requests and their frequency?

Perhaps you should target a specific request? For example, the request requesting charging values?

If this request is received within the allotted time, the RS485 link is considered active.

Then your 'Inverter Heartbeat' function measures the time between two same requests, regardless of whether the link is active or not.

When you have done all this, I will test your component with my Deye, before August 22, the date of my departure to Portugal.

Fahmula:
I’ve gone ahead and made the changes we talked about. The RS485 component now updates inverter_com_status on its own so it’s always reliable, while the heartbeat switch just controls whether the heartbeat interval (based on command 63 requests) is published. Timeout handling will flip inverter_com_status back to false if no valid 63 requests come through in time.

Can you take a look and confirm this lines up with what you had in mind?

Fahmula:
You’re right I misunderstood the purpose of `rs485_1_status.` It should indeed represent the RS485 connection status directly, with `inverter_com_status` simply mirroring that value so it remains consistent across the dashboard and AtomS3 screen. I’ll adjust the YAML accordingly.

Good point as well about the multiple RS485 inverters. I hadn’t considered that use case, but I’ll review how to structure the YAML so it supports multiple RS485 links without conflicts.

Fahmula:
Thanks for the feedback, It helped clarify the direction.

I see now that `rs485_1_status` should be the main sensor updated directly by the RS485 component, and `inverter_com_status` should simply mirror it in YAML so the dashboard and AtomS3 screen can use the same entity. That makes sense and I’ll adjust accordingly.

On the multiple‑inverter point, I think the issue isn’t with the component itself but with the YAML proxy sensors. Right now, the proxy IDs (e.g.` yambms_proxy_soc`) are hardcoded, which causes collisions if the RS485 package is imported more than once. A straightforward fix would be to prefix all proxy IDs with ${rs485_id}, so you’d end up with unique IDs like` rs485_1_proxy_soc` and `rs485_2_proxy_soc`. Then each RS485 block can link to its own set of proxies without conflicts.

Before I move forward with those changes, I just wanted to confirm with you:

Are you okay with the approach of letting `rs485_X_status` be the one updated in C++ while `inverter_com_status` simply mirrors it?

And for the multiple‑inverter case, do you agree that prefixing proxy IDs per RS485 instance is the right path, or do you see a better approach?

Lastly, regarding the Inverter Heartbeat Monitoring switch: do you think we should keep it as a diagnostic/debug feature, or should the heartbeat always be reported without needing the switch?

Example of mutiple inverter link.

 ```
rs485_pylon_1:
     ref: main
     refresh: 0s
     files:
       - path: 'packages/yambms/yambms_rs485_pylon.yaml'
         vars:
           rs485_id: 'rs485_1'
           rs485_name: 'RS485 1'
           rs485_uart_id: 'uart_esp_1’ # UART interface to which your inverter is connected
           rs485_uart_baud_rate: '9600'
           rs485_update_timeout: '5s'

   rs485_pylon_2:
     ref: main
     refresh: 0s
     files:
       - path: 'packages/yambms/yambms_rs485_pylon.yaml'
         vars:
           rs485_id: 'rs485_2’
           rs485_name: 'RS485 2’
           rs485_uart_id: 'uart_esp_2’ # UART interface to which your inverter is connected
           rs485_uart_baud_rate: '9600'
           rs485_update_timeout: '5s'
```

**Comment by Fahmula** [comment]
Question about `inverter_com_status` with multiple RS485 links

I noticed that with multiple RS485 inverters, each one would update `inverter_com_status`. This could create conflicts where one inverter reports true and another false, depending on their state and update timing.

What do you think is the intended behavior for `inverter_com_status` in this scenario?

Option A: Aggregate with OR logic, show online if any inverter is online, and only offline if all are offline.

Option B: Strict, show offline if any inverter goes offline.

Option C: Another approach (e.g., designate a primary inverter).

This way we can make sure the global status reflects the purpose you had in mind.

**Comment by Sleeper85** [comment]
It's true, I've already thought about it too but for now I don't know anyone who's ever set up two CANBUS links or a mix of CANBUS and RS485. But it will happen one day.

I'll think about this question, in the meantime, we can simply update `inverter_com_status`.

**Comment by Sleeper85** [comment]
The PYLON RS485 protocol works partially with Deye (Battery = Lithium 12).

Here are the issues I've noticed:

1. The charge and discharge amps have an extra 0 at the end (my request is 120A)
2. `rs485_status` updates to `ON` but not to `OFF` if I disconnect the connection.

Here are the changes made to the YAML for testing:

```YAML
binary_sensor:
  # --- Inverter Communication Status ---
  - platform: template
    id: rs485_status
    name: "${name} ${rs485_name} Status"
    entity_category: diagnostic

interval:
  - interval: 100ms
    then:
      - if:
          condition:
            lambda: 'return id(${rs485_id}_protocol).state == "Pylontech"  && (id(${yambms_id}_bms_combined).state > 0);'
          then:
            - lambda: !lambda |-
                // ...
                // --- Inverter Communication Status ---
                if (id(rs485_status).state)
                  id(inverter_com_status).publish_state(true);     // Set Inverter com. status to ON
                else
                  id(inverter_com_status).publish_state(false);    // Set Inverter com. status to OFF
```

image

image

Fahmula:
Thanks for testing the component! I just push the commit to the component to fix the ` rs485_status` which now switches to off once communication is lost. I'll look into the charge and discharge current now.!

Sleeper85:
Okay, everything's working properly now.

I'll test the charging current limitation tomorrow when the sun's out.

Good job.

Fahmula:
Awesome, thanks for letting me know!

This protocol was a real puzzle, but I'm happy I stuck with it. Being able to sniff the communication on my own inverter setup was the breakthrough I needed. Really appreciate you helping test it out on your end.

Fahmula:
That's great news about the 30A grid charging test, thank you for confirming!

And yes, absolutely. I will link the existing Requested Force Charge binary sensor to Bit 5 ("charge immediately") in the Pylontech 0x63 command.

While reviewing the protocol for this, I noticed another potentially useful flag: Bit 4 ("full charge request"). It seems this bit tells the inverter to charge the battery all the way to 100%. 

This brings up a related question for me regarding the existing "Enable Charge" and "Enable Discharge" switches in yam-bms. The Pylontech protocol also has bits for these:

- Bit 7: Charge Enable
- Bit 6: Discharge Enable

Should I also link your existing switches to these bits? My main question is whether those switches are intended to communicate with the inverter over the bus, or if they are just for internal logic within yam-bms. I haven't looked into them.

Knowing this will help me ensure my component integrates correctly without duplicating functionality.
Thanks again for the clear direction.

Sleeper85:
> This brings up a related question for me regarding the existing "Enable Charge" and "Enable Discharge" switches in YamBMS. The Pylontech protocol also has bits for these:
> 
>     * Bit 7: Charge Enable
>     * Bit 6: Discharge Enable
> 
> Should I also link your existing switches to these bits? My main question is whether those switches are intended to communicate with the inverter over the bus, or if they are just for internal logic within YamBMS. I haven't looked into them.

No, this is not the case with CAN bus.
These are just general YamBMS switches that I never use anyway.

Fahmula:
Alright I'll work on the force charge right now so you can test it. Not sure if my inverter will respect it.

Fahmula:
I've implemented the Requested Force Charge feature. The component now sets Bit 5 in the Pylontech 0x63 command whenever the sensor is active, which should be the correct instruction.

My own inverter doesn't seem to react to this command as I expected. I'm very keen to hear if it works on your setup. Let me know how it goes when you get a chance to test it.

Sleeper85:
This works perfectly with Deye.

Thanks again for this great integration.

I'm merging this.

PR #77, opened 2025-08-05
fix(yambms_rs485_pylon.yaml): Prevent stale data on BMS disconnect

Fahmula:
Previously, the data-forwarding logic would only execute if a BMS was connected. When the BMS disconnected, the logic would halt, leaving the proxy sensors with their last known values. This caused stale data to be sent to the inverter.

This change refactors the YAML logic to explicitly handle the BMS offline state.

When a BMS is no longer available, all numeric proxy sensors are now invalidated by publishing `NAN`, and all binary alarm/protection sensors are reset to `false`. This ensures the C++ component receives a clear signal that the data is no longer valid, allowing its fail-safe mechanisms to engage and stop communication.

This fix improves overall system robustness by preventing stale state and ensuring a predictable response to a BMS disconnect.

### DIFF
--- a/packages/yambms/yambms_rs485_pylon.yaml
+++ b/packages/yambms/yambms_rs485_pylon.yaml
@@ -159,50 +159,86 @@ interval:
     then:
       - if:
           condition:
-            lambda: 'return id(${rs485_id}_protocol).state == "Pylontech"  && (id(${yambms_id}_bms_combined).state > 0);'
+            lambda: 'return id(${rs485_id}_protocol).state == "Pylontech";'
           then:
             - lambda: !lambda |-
-                // --- PROTOCOL ENABLED: Pass real data if available ---
-                // The .has_state() check prevents race conditions at startup.
-                // Mirror RS485 status into inverter_com_status
-                if (id(${rs485_id}_status).has_state()) id(inverter_com_status).publish_state(id(${rs485_id}_status).state);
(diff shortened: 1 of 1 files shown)


Sleeper85:
This seems to work, but the `inverter_com_status` is not updated.

The data received by the inverter is all 0.

What I want to be sure of is that your component is no longer responding to requests from the inverter when there are no combined BMSs.

The goal is to avoid sending it false information, and you shouldn't send it zero values either.

In fact, you simply need to stop responding to requests from the inverter.

Is this the case ?

Fahmula:
You are absolutely correct. Thank you for the detailed feedback. Your testing uncovered a subtle race condition in my C++ component.

What was happening: When the BMS disconnected, the component would receive an inverter request and send one last packet of zeros before its main loop could process the state change.

The Fix: I will implemented two fixes:

- C++ Component: I will moved the data validity check from the background loop() directly into the route_frame_request_() function. This guarantees that the component checks for NAN values at the exact moment of a request. If it finds them, it will now immediately and completely stop responding, which is the correct behavior you described.
- YAML: I will also add a line to update the inverter_com_status when the BMS disconnects.

With both of these changes, the system should now be perfectly robust. Thank you again for helping me with finding these bugs.

Fahmula:
I have pushed commits that should fix both issues. Can you test to confirm?

PR #79, opened 2025-08-07
switch to ledc and monochromatic light for a smoother light effect

shvmm:
With the default 1 second light transition time coincides with 1 second 0x305 CAN ACK it looks good. Thats all

### DIFF
--- a/packages/board/board_ESP32_DevKit-V1.yaml
+++ b/packages/board/board_ESP32_DevKit-V1.yaml
@@ -69,13 +69,13 @@ logger:
 light:
   # ESP Light used to see the inverter heartbeat
   # Internal : only specifying an id without a name will implicitly set this to true.
-  - platform: binary
+  - platform: monochromatic
     output: esp_led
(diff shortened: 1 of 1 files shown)


Sleeper85:
Thanks for this PR but it doesn't work well with `PYLON 1.2` and an `ACK` every `600ms`.
The LED doesn't flash.

It's fine with `PYLON V2`.

shvmm:
> Thanks for this PR but it doesn't work well with `PYLON 1.2` and an `ACK` every `600ms`. The LED doesn't flash.
> 
> It's fine with `PYLON V2`.

Was just meant as an optional `nice-to-have`
BTW, the default light transition length can updated to suit PYLON 1.2. 
I suggest 0.5s

Sleeper85:
Thanks for the clarification.

Unfortunately, I can't predict in advance which protocol users will choose, so I think it would be simpler to keep the current version without any transition.

Sorry

PR #82, opened 2025-08-11
Add support for external balancer

GHswitt:
WIP

This adds support for external balancers (using JK BLE), see also: https://github.com/Sleeper85/esphome-yambms/discussions/78


Sleeper85:
I tried the `!extend` to add filters to the BMS sensor so I can centralize the code in a single YAML but it doesn't work with JK-BMS sensors (UART).

```YAML
sensor:
  - id: !extend bms${bms_id}_balance_trigger_voltage
    filters:
      - lambda: |-
          if (id(bms${bms_id}_var_ext_balancer_online))
            return id(bms${bms_id}_var_ext_balancer_trigger_voltage);
          else
            return id(bms${bms_id}_balance_trigger_voltage).state;
```

GHswitt:
First test looks ok:

image

However it seems the balancing entities need to be renamed:

image

Sleeper85:
> I added the JK only because the support comes from esphome-jk-bms. The balancers, Heltec/NEEY seem to belong to the same hardware family so the idea was: This file supports balancers that are supported by esphome-jk-bms via BLE.
> 
> But as you said, I think for other people it would be easier to find the right file if the name is in it.
Okay, for me, it's actually more complicated to understand how it works with NEEY.

But it's true that it's the JK BLE protocol...

It'll have to be explained in the documentation.

GHswitt:
> Would it be better to include `bms_balancer.yaml` directly in `bms_base.yaml`?
> 
> What do you think?
> 
> Thanks again for your work.

Yes, as bms_balancer.yaml will be included in all BMS we could just put it in bms_base.yaml.

You're welcome, thanks for your work :-)

Sleeper85:
> > Why did you remove the sensor names?
> > We need to keep them and just change the ID.
> > image
> 
> Because some sensor had the same name like the one in `bms_balancer.yaml`:
> 
> ```
> binary_sensor:
>   # equalizing
>   - platform: template
>     id: bms${bms_id}_equalizing
>     name: "${name} ${bms_name} equalizing"
> ```
> 
> And then I got a compile error that the name was already used.
> 
> So how should the BMS sensor be named? `name: "${name} ${bms_name} balancing"`? Then I would try to check all are using the same name.

We could keep the original name and add the `internal: true` option.

This should compile fine.

GHswitt:
I've kept `bms_balancer.yaml` because `bms_combine_modbus_client.yaml` needs it. 
But now it is included in `bms_base.yaml` and not in each BMS file seperately.

Sleeper85:
Your example `RP_example_multi-node_YamBMS_modbus_server_3.yaml` contains the BMS already present in `server_2`.
Please remove BMS from `server_3`.

Sleeper85:
Shouldn't these substitutions in `balancer_modbus_client.yaml` be named `balancer`?

```YAML
substitutions:
  bms_model: "YamBMS"
  bms_protocol: "Modbus"
```
Like this:

```YAML
substitutions:
  balancer_model: "JK NEEY"
  balancer_protocol: "YamBMS Modbus"
```

In `balancer_sensors_JK_NEEY_BLE.yaml`, rename like this:

```YAML
substitutions:
  balancer_model: "JK NEEY"
  balancer_protocol: "BLE"
```

Sleeper85:
For all `bms` YAMLs, please add `internal: true` for the `balancing` and `balance_trigger_voltage` sensors.

The two new sensors are sufficient and will display either the `BMS or Balancer` values.

GHswitt:
Thanks for the review, I think I got everything...

Sleeper85:
In YamBMS `LP` and `RP` multi-node example.

image

Sleeper85:
> Shouldn't these substitutions in `balancer_modbus_client.yaml` be named `balancer`?
> 
> ```yaml
> substitutions:
>   bms_model: "YamBMS"
>   bms_protocol: "Modbus"
> ```
> 
> Like this:
> 
> ```yaml
> substitutions:
>   balancer_model: "YamBMS"
>   balancer_protocol: "Modbus"
> ```
> 
> In `balancer_sensors_JK_NEEY_BLE.yaml`, rename like this:
> 
> ```yaml
> substitutions:
>   balancer_model: "JK NEEY"
>   balancer_protocol: "BLE"
> ```

Also in `balancer_modbus_JK_NEEY_BLE.yaml`

```YAML
substitutions:
  balancer_model: "JK NEEY"
  balancer_protocol: "BLE-Modbus"
```

Sleeper85:
Please preserve the modified sensor origin names for the following YAMLs:

* packages/bms/bms_combine_JK_RS485_Modbus_bms_minimal.yaml
* packages/bms/bms_combine_JK_UART_4G-GPS_minimal.yaml
* packages/bms/bms_sensors_JK_RS485_Modbus_JK-B_standard.yaml
* packages/bms/bms_sensors_JK_RS485_Modbus_bms_full.yaml
* packages/bms/bms_sensors_JK_RS485_Modbus_bms_standard.yaml
* packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml


**For SEPLOS V3, like this:**
* packages/bms/bms_sensors_SEPLOS_V3_RS485_bms_full.yaml

```YAML
    id: bms${bms_id}_bms_equalizing
    name: "${name} ${bms_name} Equilibrium module to open"
    internal: true

    id: bms${bms_id}_bms_balance_trigger_voltage
    name: "${name} ${bms_name} Equalization open voltage difference"
    internal: true
```

Sleeper85:
This note is missing from the `YamBMS_RP_YAML.md` version of the readme.

image

GHswitt:
I added a picture marking the 120 Ohm resistor for the T-CAN485. For the two other LilyGo boards I could not yet find good pictures to mark them.

Sleeper85:
Please replace the GitHub address at the beginning of all your BMS BASEN YAML files with your own: https://github.com/GHswitt/esphome-basen

Also for **bms_modbus_DEYE_CAN_module_full.yaml**

GHswitt:
Done

Sleeper85:
Yes, but it's done, thanks to you too.

I'm going on vacation for a month in a week.

I'll check your `Auto EOC` PR, but I won't be able to test anything, so I hope you've tested this feature yourself.

Sleeper85:
We've internalized some sensors that were previously "number" and are no longer editable (JK BLE YAML).

I'll fix that...

image

PR #85, opened 2025-08-15
Adding EG4 LLV2 BMS

RAR:


RAR:
> Hi,
> 
> Thanks for this PR.
> 
> Can you add a BMS import example `LP` and `RP` to the `examples/single-node` folder ?

I've added them, however, its made me realize I should separate out yambms bms_id and the eg4 modbus address.  EG4 has an odd thing where if you have the ID on the BMS set to 1, it completely changes the modbus output, and they've decided not to document it.  

Andrew

Sleeper85:
The first combined BMS is the one with `bms_id: 1`.

It shouldn't work if you start at `2`.

It is therefore necessary to configure a separate `modbus_id`.

GHswitt:
Nice work! I would just recommend to rebase/merge the commits to reduce the number :-)

Sleeper85:
`bms_id` is not the Modbus address.
Your comments in the example YAML are confusing.
Could you please adapt the comments ?

```YAML
  bms1: !include
    file: packages/bms/bms_combine_EG4_LLV2_RS485_bms_full.yaml
    vars:
      bms_id: '1' # BMS id and modbus address
      bms_name: 'BMS 1'
      ## BMS address must start at 2 with EG4 LLV2, ID 1 is a special case and has different output.
      bms_address: 2
```

RAR:
Yes, that is very confusing, I've updated them to match the JK-B comments.  

As I'm sure you have other formatting and header comments you'd like to set - if you'd like you can grab these files, change them as you like and commit them directly - no need to actually use this PR if you don't want to, this was just an easy way to get them over to you.  

Andrew

PR #111, opened 2025-11-21
Add ability to disable periodic version check and new FLIP_C3 board

5hortCircuit:
### Added the ability to disable the periodic version check
The logic was already in the `yambms_service.yaml` but was not exposed to the upper levels.  Added a variable called `periodic_version_check` that is mapped back to the `{yambms_id}_service` that is checked as part of the service before it goes out to the internet.  Also added a INFO print when the service is disabled so that the user can see (if watching logs) that the check was disabled or that their network was down.

### Added a new board configuration for the FLIP_C3
The FLIP_C3 is a simple ESP32-C3 board that has an integrated power supply that has a 60V input.  This board can be connected directly to a 48V battery and then be used with the M5Stack Isolated RS485 Unit (U094) to get a RS485 connection back to the master unit.  I successfully used 4 of these units to communicate with the JK-BMS B-Series BMS's.  The FLIP_C3 has an onboard 5V regulator that can power the RS485 module, 2 UART exposed on connectors that can be used for TTL to the BMS and TTL to the RS485 module. In the US it is available on Amazon, product page: https://wiki.vdbx.io/product/flip_c3

### DIFF
--- a/YamBMS_Local_Packages_example.yaml
+++ b/YamBMS_Local_Packages_example.yaml
@@ -216,7 +216,10 @@ packages:
   # | >>> YamBMS
   # +------------------------------------------------------------------------------+
 
-  yambms: !include packages/yambms/yambms.yaml
+  yambms: !include
+    file: packages/yambms/yambms.yaml
+    #vars:
+    #  periodic_version_check: 'true'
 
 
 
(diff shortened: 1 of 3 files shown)


5hortCircuit:
Closing to make changes to the `dev` branch.  Making separate PRs for adding the board and the other chances.

PR #117, opened 2025-11-25
Round requested charge/discharge current to nearest integer

GHswitt:
Because of temperature based current limitation, the requested charge/discharge current could be non-integer values. For the used high-power applications, sub 1A precision is not required and may add noise.

Before:

image

After:

image

### DIFF
--- a/packages/yambms/yambms_combine.yaml
+++ b/packages/yambms/yambms_combine.yaml
@@ -314,7 +314,7 @@ interval:
                 max_charge_current = rate_max_charge_current;
             }
             
-            id(${yambms_id}_max_charge_current).publish_state(max_charge_current);
+            id(${yambms_id}_max_charge_current).publish_state(round(max_charge_current));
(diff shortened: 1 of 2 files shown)


Sleeper85:
Thank you for this optimization ;-)

GHswitt:
@MrPabloUK: Thanks for the information. For the requested charge/discharge current or the actual current measured by the BMS?

Even with the protocol allowing it, I don't think that many inverters are able to govern the charge/discharge current with 0.1A precision (or even 500mA). With the current ranges involved that would go into the range of high-end lab equipment. Which is much more expensive. 100mA at 52.8V is just about 5.2W. When I have a range of 10kW that would be 0,052%.

But if there is someone using this precision (and maybe has an inverter datasheet stating the precision it can use) I'm open to make this configurable.

Update: :-) Ok, saw that too late. Yes, it was full integers.

PR #119, opened 2025-11-27
Update YamBMS_HA_Dashboard_JK_RS485_7xBMS.yaml

LocoSlug:
Added some segments to the gauges

Screenshot 2025-11-27 at 1 52 20 AM


Sleeper85:
I'm not sure if choosing a default value of **12000W** is a good idea.
I think each user will need to update their dashboard to match their specific setup.

That said, I also realize that the values ​​I chose a few years ago aren't necessarily better either.

We should almost conduct a survey to find the most suitable default values.

It would also be necessary to align the amps and watts :

* 12000W / 250A
* 6000W / 125A

Something to think about...

LocoSlug:
Totally understand, choosing default values is tricky. 

Any thoughts on the needle segments having discharge to the left and charging to the right?

PR #130, opened 2025-12-08
[EG4] Swap away from esphome modbus component

RAR:
Swaps to using a C based implementation rather than using the built in modbus component in esphome.

https://github.com/RAR/esphome-eg4-bms


RAR:
My apologies, I'm not sure how I missed that!  

I've corrected it in [v1.0.1](https://github.com/RAR/esphome-eg4-bms/releases/tag/v1.0.1)

Let me get a fresh PR put together here.

PR #134, opened 2025-12-12
Sub device support

MrPabloUK:
Added sub device support to yambms, no additional configuration required by users as most sub device names are automatically inherited.

image

Please note that the following sub-devices have a hard-coded name, as there is no name variable in the main yaml config.
- Modbus sniffer


MrPabloUK:
Ok, let me know when you've pushed all your changes to Dev and I'll resolve any conflicts as needed.

Sleeper85:
I think you can fix the EG4 BMS.
I'll probably merge this branch into main soon because there are too many changes now.

I'll fix my own version after.

PR #147, opened 2025-12-31
CCL, DCL and Auto Temp Current Control

MrPabloUK:
**Temperature Derating**

This has been moved out of yambms_combine and into a dedicated package file, similar to other auto functions.
~The original substitution used to map temperatures to C-rates has been deprecated, replaced by a calibrate_linear filter on the relevant temperature sensor.~
The adjusted C-rate is then applied to total battery capacity, an offset from the configured max current calculated and stored.
~If we need to support a different temperature to C-rate mapping (for a different chemistry, etc), then we may need to have several auto_temp packages that are selected as required. This is a _future problem_.~

The temperature -> C-rate mapping has been converted into a table, allowing for varying numbers of data points.

**Requested CCL and DCL logic changes**

In the last version of code, temperature derating, EOC derating and autoCCL derating were applied at the same time - this resulted in an incorrectly large reduction in requested CCL.
Now, all derating values are compared and the most strict one is applied.
For example, temperature based derating may be in effect, but then increasing cell voltages means that AutoCCL becomes the limiting factor.

The same approach has been applied to Requested DCL, although there are less derating functions in place here.

**Derating Reason Sensors**

There are several functions that can reduce requested current, so these two new sensors explicitly state which function is limiting current (if any).


Sleeper85:
Thanks for this PR.

The idea behind allowing users to enter **temp_c-rate** values ​​into the substitution variables was to make it user-configurable. The current values ​​are those from the EVE datasheets. Most users will probably keep them as default, but it would be nice to be able to make them configurable.

I'm currently abroad and can't test anything.

I'm happy to validate this PR so @GHswitt can test it.

MrPabloUK:
That's what I thought, and I tried several ways to keep a substitute or similar approach.
The only other option is a much more complex lambda, and I'd prefer not to do that as it's more maintenance headache if changes are needed.

MrPabloUK:
I'm very interested to see the v2 code, I haven't developed any native components before so it'll be a good learning experience.
Will you be adding it to the dev branch once 1.6 is released?

Sleeper85:
> I'm very interested to see the v2 code, I haven't developed any native components before so it'll be a good learning experience. Will you be adding it to the dev branch once 1.6 is released?

Yes, as soon as 1.6.0 is released and after adapting it to reflect the latest changes.

MrPabloUK:
I have a lambda based solution that will work well with a substitution table, just got to test it today and then I'll update the PR.

MrPabloUK:
@Sleeper85, if you're converting this into an ES phone component, should all auto functions be 100% c++ logic?

Sleeper85:
Not necessarily, but it would be better in order to be able to interconnect directly with the shared YamBMS structure.

MrPabloUK:
Well, that's an annoying error that I made regarding the cell temperature. I've corrected that.
It made me realise that I'll need to enhance the logic to reference max temperature as well for high temp derating.
What's your total battery capacity?

Sleeper85:
Perhaps I could validate this PR as it is now, even though it's not possible for the user to define the c-rate themselves.

I think it's time to replace version **1.5.8** with **1.6.0**.

The rest could be done in a second step for 1.6.1 or in V2.

@MrPabloUK if you think you'll have some time in the coming days to develop a function in a lambda, I can wait.

I'd like to release 1.6.0 **before January 22nd** because I'll have more time after January 22nd to continue working on the version with esphome components.

MrPabloUK:
I actually just pushed one last change to my testing setup which has worked well.
I'll update this PR today, it is a lot more effective now plus has user defined c-rates.
I'll also add documentation for this feature in the readme directory.

Sleeper85:
Thanks, good job 😉

MrPabloUK:
I have c++ based code for AutoCCL and AutoDCL that I'll be testing soon.
I'll hold off on any PR for this until the v2 branch is available.

Sleeper85:
There's a problem...

At -24°C, the requested charging and discharging current should be 0A.

image

Sleeper85:
Below 0°C, the charging current must be 0A.
From 0°C, it can be 0.05A.

At -30°C or below, the discharging current must be 0A.
Above -30°C, it can be 0.5A.

At 60A or above, the current must be 0A in both cases.

That's what I understand from reading the EVE datasheet.

For reference, here are the values ​​from the datasheets of the EVE LF280K V3 and MB31 cells (1P = 1C).

Image

Image

Sleeper85:
@GHswitt 

Can you confirm that the current version **1.6.0** works well for you ?

MrPabloUK:
@Sleeper85, I see you have updated the charge table but the discharge table also needs to be updated.

Sleeper85:
I tested it myself too and it looks good 😉

MrPabloUK:
@Sleeper85, what is the refresh interval set to in your test configuration?

Issue #145, opened 2025-12-30
Temperature-based current limitation fluctuations


GHswitt:
Hi,

currently it is quite cold and I've enabled Temperature-based current limitation.

The minimum temperature fluctuates, leading to many changes in the requested charge current:

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/8cda55ba-2ac7-4ff9-adaf-5263cc0018f4" />

Where is the best place for a filter? Minimum temperature in general? Or filter it at input of the temperature-based current limitation?  Or filter output value of the temperature-based current limitation?

Best regards,
Sebastian

Sleeper85:
I don't know how to handle it, what I think at the moment is that a decrease in the required current should not be delayed and at the same time this unnecessary fluctuation must be eliminated.

GHswitt:
i can test some solutions when I'm back on the 2.1. I tend to filter the temperature a bit on the input of the calculation and linearize the output value between the _var_charging_rate points as @MrPabloUK suggested.

Sleeper85:
> I tend to filter the temperature a bit on the input of the calculation and linearize the output value. 

Yes also for the temperature. 

On the graph, the temperature drops directly from 5 to 4 degrees. There are no decimal places?

Sleeper85:
@MrPabloUK 

There's no rush, we're in the development branch and we're in the continuation of the end-of-year festivities and family time.

GHswitt:
Yes, no rush :-) Currently is slow time between the years. If something takes several weeks, that is fine for me. I like to test longer to make sure it works like intended.

Sleeper85:
@MrPabloUK thanks for your PR 😉

@GHswitt Feel free to share the results of your upcoming tests and improvements here 😉

GHswitt:
Done, I've modified the filters a little and now I'm happy, no more noise:

<img width="3441" height="1440" alt="Image" src="https://github.com/user-attachments/assets/4a016f4b-ee40-4a50-b268-c4f2ab04e449" />

I will create PR with the changes.

GHswitt:
Closing, fixed with PR https://github.com/Sleeper85/esphome-yambms/pull/155
Thanks for your support.

PR #155, opened 2026-01-11
Update temperature sensor filters to reduce noise

GHswitt:
These commits update the filters for the temperature sensors to reduce noise. See https://github.com/Sleeper85/esphome-yambms/issues/145 for before/after graphs.


GHswitt:
Done.

PR #158, opened 2026-01-15
Automatically disable shunt when connected BMSs are going offline

GHswitt:
This PR adds the possibility to provide connected BMS IDs to a shunt.

If one or more BMS from this array are offline, the shunt is not used anymore (Combine = Disabled). This avoids returning a wrong SOC when not all batteries are online/available anymore.

The `bms_ids` variable is used provide an array of BMS IDs that are connected to a shunt.

This variable is optional. If not provided, the shunt works independently from the BMS online status like before.

Examples:
```
bms_ids: '{1}'
bms_ids: '{1, 2}'
bms_ids: '{4, 5, 6}'
```

Additionally a diagnostic text_sensor has been added to show the cause of the switch change, this hopefully avoids confusion why the switch was disabled automatically:

image

If the switch is turned on:

image

If the switch is turned off:

image

### DIFF
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.01.14
-# Version : 1.6.0
+# Updated : 2026.01.24
+# Version : 1.6.1
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -926,4 +926,6 @@ views:
                 name: Shunt 1 Combine enabled
               - entity: binary_sensor.shunt_1_can_be_combined
                 name: Shunt 1 Can be combined
+              - entity: sensor.shunt_1_state
+                name: Shunt 1 State
         column_span: 1
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_JK_RS485_3xBMS.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_JK_RS485_3xBMS.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.01.14
-# Version : 1.6.0
+# Updated : 2026.01.24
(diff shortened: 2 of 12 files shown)


GHswitt:
I've added a second commit with which I'm not that happy.

Original behavior: Disable shunt usage if assigned BMS goes offline, remember that we did it. Check if it is online again and enable shunt usage if we did disable it before.

This is fine for short interrupts. What about longer times? So I would like to only enable it again when the difference between BMS SOC and shunt SOC is small (configurable, default 10%).

However for that I need the BMS SOC and I added an additional array and I do not really like that solution, but substitutions limit the available options.

So it is either that or only disable shunt combine, without automatic enabling when it is back online. Enabling is then left to the user. This would make the code much shorter and easier.

Sleeper85:
Did you know that the `on_turn_on` / `on_turn_off` of a switch isn't executed during boot ?
If you want it to be executed, use a lambda function within an `on_state` statement.

```
switch:
  # Combine enable/disable Switch
  - platform: template
    name: "Combine enabled"
    id: shunt${shunt_id}_switch_combine
    device_id: shunt_${shunt_id}
    restore_mode: RESTORE_DEFAULT_ON
    optimistic: true
    on_turn_on:
      - lambda: |-
          id(shunt${shunt_id}_bms_caused_uncombine) = 0;
          id(shunt${shunt_id}_combine_state).publish_state("Enabled (switch)");
    on_turn_off:
      - lambda: |-
          if (id(shunt${shunt_id}_bms_caused_uncombine))
          {
            std::string state_msg = "Disabled (BMS " + std::to_string(id(shunt${shunt_id}_bms_caused_uncombine)) + " offline)";
            id(shunt${shunt_id}_combine_state).publish_state(state_msg.c_str());
          }
          else
          {
            id(shunt${shunt_id}_combine_state).publish_state("Disabled (switch)");
          }
```

GHswitt:
I don't see why it should be executed during boot. I checked the states after reboot and they were the same.

Of course, the reason gets lost, for that `_bms_caused_uncombine` could be preserved but I don't want to do that. The past values can still be seen/found in HA history when required.

Sleeper85:
It was just to be sure of the behavior. If it works well, that's perfect.

PR #162, opened 2026-01-19
Merge PACE Sensor Fixes

bradmatt275:
These are some fixes to issues I found while testing the new RS485 PACE configurations.

Most of them are due to the PACE BMS reporting the values in mV rather than V.

* Add PACE-specific BMS configuration and battery SOC handling
* Refactor Battery SoC and Capacity Remaining calculations to use direct values from PACE BMS
* Refactor PACE BMS configuration to include new balancer file and update voltage handling for SOC calculations
* Fixed online sensor showing as unavailable

### DIFF
--- a/packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
+++ b/packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
@@ -1038,7 +1038,8 @@ number:
     address: 65
     register_type: holding
     value_type: U_WORD
-    multiply: 0.001
+    lambda: "return x * 0.001f;"
+    write_lambda: "return x * 1000.0f;"
     min_value: 0.0
     max_value: 65.535
     step: 0.001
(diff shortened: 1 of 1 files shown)


Sleeper85:
Thank you for your PR, but this amounts to creating a custom version specifically for PACE and even more unnecessary YAML to maintain.

The only thing that needs to be done is to modify the original PACE sensors so they report **voltages (V)** instead of **millivolts (mV)**. Only the YAML file `bms_sensors_PACE_RS485_bms_full.yaml` needs to be adapted so they can rely on other YAML files like all other BMS.

Please also keep the `ID names` like the other BMS so that the code remains consistent.

Thank you for understanding.

Sleeper85:
For the `number`, add an `on_value` which will update two new `template sensor` ​​expressed in `V` for the following **id** ​​`bms${bms_id}_cell_ovp` and `bms${bms_id}_cell_uvp`.

If you deem it necessary, of course.

bradmatt275:
@Sleeper85  No problem I completely understand the need to keep it standard. I have just updated and tested it from my end side:

image

image

image

PR #167, opened 2026-01-21
Service get JSON - Resolves 2026.1 incompatibility

MrPabloUK:
The error below is due to a change in AnduinoJSON as part of the latest update to ESPhome 2026.1.
```
/data/packages/4fe5b35a/packages/yambms/yambms_service.yaml: In lambda function:/data/packages/4fe5b35a/packages/yambms/yambms_service.yaml:64:44: error: call of overloaded 'publish_state(ArduinoJson::V742HB22::detail::MemberProxy<ArduinoJson::V742HB22::JsonObject, ArduinoJson::V742HB22::detail::RamString>)' is ambiguous 64 | id(${yambms_id}_last_version).publish_state(root["version"]); | ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~In file included from src/esphome/core/application.h:47, from src/esphome/components/api/api_frame_helper.h:13, from src/esphome/components/api/api_connection.h:5, from src/esphome.h:3, from src/main.cpp:3:src/esphome/components/text_sensor/text_sensor.h:44:8: note: candidate: 'void esphome::text_sensor::TextSensor::publish_state(const std::string&)' 44 | void publish_state(const std::string &state); | ^~~~~~~~~~~~~src/esphome/components/text_sensor/text_sensor.h:45:8: note: candidate: 'void esphome::text_sensor::TextSensor::publish_state(const char*)' 45 | void publish_state(const char *state); | ^~~~~~~~~~~~~
```

This small change explicitly casts it as a string, resolving the issue.
This change should be backwards compatible as well.


MrPabloUK:
Yes, I mistakenly used main as the base for my branch, then pulled in dev changes afterwards.

PR #170, opened 2026-01-24
Add Freenove S3 board MCP2515 defaults and clean substitutions

txubelaxu:
## Summary
- add default "canbus_clock" substitution for MCP2515 to avoid invalid example builds
- clean up and align the Freenove ESP32-S3 board MCP2515 substitutions
- update board metadata (date/version, board_name)

## Notes
- MCP2515 clock defaults to 8MHz; users can override to 16MHz when their board uses a 16MHz crystal
- MCP2515 pins are kept contiguous and in board order to match common flat cables

### DIFF
--- a/packages/board/board_ESP32-S3_freenove_esp32_s3_wroom.yaml
+++ b/packages/board/board_ESP32-S3_freenove_esp32_s3_wroom.yaml
@@ -1,0 +1,88 @@
+# Updated : 2026.01.24
+# Version : 1.1.1
+# GitHub  : https://github.com/Sleeper85/esphome-yambms
+
+# YamBMS ( Yet another multi-BMS Merging Solution )
+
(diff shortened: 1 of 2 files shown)


Sleeper85:
> clock (Optional, enum): The frequency of the clock crystal used on the MCP2515 device. One of 8MHZ, 12MHZ, 16MHZ or 20MHZ. Defaults to 8MHZ.

Thank you for your PR.
To be perfectly clear, the `substitutions` should be named `mcp2515_clock`.

As soon as this is changed, I will merge your PR.

Sleeper85:
I renamed the YAML file to `board_ESP32-S3_Freenove.yaml`

I assume the version you are using is probably the `Lite` version, but in any case the GPIOs are the same as the versions allowing the addition of a screen.

PR #172, opened 2026-01-24
Gate Victron CAN frames until values are valid at boot

txubelaxu:
## Summary
- prevent early Victron CAN frames with zeroed values during ESP32 boot
- only send 0x351/0x355 when required data is available (Victron only)
- keep behavior unchanged for non-Victron protocols

## Files touched
- packages/yambms/yambms_canbus.yaml
- related device/package YAML updates for default values and docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,0 @@
-# Gitignore settings for ESPHome
-# This is an example and may include too much for your use-case.
-# You can modify this file to suit your needs.
-/.esphome/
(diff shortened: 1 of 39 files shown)


Sleeper85:
Regarding the source of the component `JK_RS485`, I'm OK to change the source to your git repository if all the changes I have made are merged on your end.
Give me some time to open a discussion on your GitHub so we can check together what needs to be done.

PR #179, opened 2026-01-29
New LED Control Logic - Improves Compatibility and Accessibility

MrPabloUK:
This pull request introduces significant updates to YamBMS led control, focusing on integrating a new status_in sensor for LED management. The changes aim to improve compatibility for non RGB board and accessibility for colour blind users.

### Status Integer Sensor
Introduced a *_status_int sensor in yambms_canbus.yaml and yambms_rs485_pylon.yaml
Determines LED behavior based on ESP32 and CAN/RS485 status.
This sensor can be added to other components as well as needed, for example I've added it to the modbus yaml used to connect to my Solis inverter.

### New LED Control Logic
Replaced the existing RGB led logic with new code that provides the following features.

4 states utilising both colour (where available) and flash codes:

| State | Level | Color | Flash |
| --- | --- | --- | --- |
0 | Critical | $${\color{red}Red}$$ | Quad flash
1 | Warning | $${\color{yellow}Yellow}$$ | Triple flash
2 | Info | $${\color{blue}Blue}$$ | Double flash
3 | Healthy | $${\color{green}Green}$$ | Single flash

Timing:
Flash ON duration: 150ms.
Flash cycle slot: 300ms (150ms ON + 150ms OFF).
Full cycle duration: 2000ms (includes a pause after the last pulse).

The LED state is updated only when there is a change in the desired state or flash timing, reducing unnecessary updates.

Customisable Substitutions:
`status_sensor`: Defines the sensor used to determine the LED behavior.
`light_id`: Specifies the LED light entity. Multiple LEDs can be controlled, for example `led_1` for RS485 status, `led_2` for CAN status.

Compilation Safe:
If no light or status sensor is provided as an input, `nullptr` is used as a default to prevent ESPhome compilation errors.
This means the led control package can always be included in the can and rs485 package, even if no led is specified in the main YAML.

### Board Configs Updated (Incl. T-Connect LEDs)

All board YAML files now have an additional substitution `board_has_rgb`.
This must be set to true or false, as it's referenced in the `board_options_led_control.yaml` file.

The T-Connect config has also been updated, with the addressable 4 leds are also referenced separately. This means a led can be dedicated to each port, making it easier to see status without using Home Assistant. The default led id is still `esp_light`, referring to all 4 LEDs as one unit.
For example, my board is in this enclosure:
image
P1 is dedicated to JK batteries, P2 for a Solis inverter via RS485, then finally P4 for battery CAN communication.
The individual LEDs show status of each of these components.

Example YAML files and documentation have been updated.

Testing has been successful on 3 different ESP32 boards, both RGB and non-RGB, but I'd welcome any other test results before this is merged.


Sleeper85:
I checked a few boards, and the YAML file `board_options_led_control.yaml` is never added as a package.

Is that normal ?

Sleeper85:
Okay, I hadn't replaced `canbus_light_id` with `light_id`.

Importing `board_options_led_control.yaml` from CAN bus or RS485 will have the disadvantage of losing the connection status with HA (ESP_OK) in the following case:

In the case of a YamBMS without a CAN bus or RS485 output (I personally use it in one of my setups solely to have a dashboard/overview of my BMS on a system with an MPPT charger without inverter communication).

Have you also considered the possibility of importing the CAN bus and/or RS485 packages multiple times simultaneously to distribute the information to several inverters of different brands?

In my opinion, the code managing the LED should be imported from the board.

MrPabloUK:
I'll create a feature improvement issue (hopefully today) where we can discuss further.
I have a few ideas that would handle multiple scenarios.

The intention is to have a LED based alert system (and potentially other outputs too) that can ingest data from one or more inputs, then output on one or more LEDs accordingly.
Multi-LED support is a must in my opinion, as the T-Connect is a recommended board and we are not using all the functionality.

Your point about using YamBMS with no inverter (or other system) output is a valid one. It also makes sense to consider how a multi-node system will report errors as well.

Let's consider this PR an initial step towards a more accessible LED alert system. It provides feature parity for non-RGB boards, plus handles accessibility concerns. We can certainly do a lot more beyond this.

MrPabloUK:
Moving the current code to the board means multi led functionality will now be unavailable / broken.
If 1.7.0 is not an urgent release, it makes more sense to focus on a true multi input, multi output status reporting system.

I have the core concepts defined, code will be completed in the next few days.

Sleeper85:
> I assume this is so the led control logic can be used in non-YamBMS projects? 
> This suggests the only point of commonality would be the board package directory?

Yes

To make possible this kind of configuration (sensors replace combine) to monitor only BMS without YamBMS, which is also necessary for `multi-node` where there is no YamBMS on the server nodes.

```
  bms:
    ref: main
    refresh: 0s
    files:
      # BMS 1
      - path: 'packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml'
        vars:
          bms_id: '1' # must be a number
          bms_name: 'JK-BMS 1'
          bms_uart_id: 'uart_esp_1'
          # Maximum cell charging cycles is used to calculate the battey SoH.
          # MB31=8000.0, LF280K v3=8000.0, LF280K v2=6000.0, LF280=3000.0 (decimal is required)
          bms_cell_max_cycles: '6000.0'
      # BMS 2
      - path: 'packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml'
        vars:
          bms_id: '2' # must be a number
          bms_name: 'JK-BMS 2'
          bms_uart_id: 'uart_esp_2'
          # Maximum cell charging cycles is used to calculate the battey SoH.
          # MB31=8000.0, LF280K v3=8000.0, LF280K v2=6000.0, LF280=3000.0 (decimal is required)
          bms_cell_max_cycles: '6000.0'
```

PR #181, opened 2026-01-30
Board Config Update - Remove Outdated Package

MrPabloUK:
Board configs still referenced the previous RGB LED yaml - this is now corrected.


Sleeper85:
I had a feeling there was a problem 🙃

PR #192, opened 2026-02-03
Report N/A SoC before BMS combine

txubelaxu:
This fix prevents YamBMS from publishing a misleading 0% SOC before any BMS has been combined. It changes the internal SOC variable to a float and publishes NAN (shown as N/A in Home Assistant) when no BMS is available. Once a valid SOC is computed, it publishes the real value as before.


txubelaxu:
Sleeper85, could you please delete this PR? 
Many thanks in advance

Sleeper85:
The PR is closed, there's nothing more we can do. If it's causing problems, you can delete the source branch (the one on your GitHub).

PR #193, opened 2026-02-03
Report N/A SoC before BMS combine

txubelaxu:
This fix prevents YamBMS from publishing a misleading 0% SOC before any BMS has been combined. It changes the internal SOC variable to a float and publishes NAN (shown as N/A in Home Assistant) when no BMS is available. Once a valid SOC is computed, it publishes the real value as before.

### DIFF
--- a/packages/yambms/yambms_core.yaml
+++ b/packages/yambms/yambms_core.yaml
@@ -389,7 +389,7 @@ interval:
             int power = 0;
             int battery_charging_power = 0;
             int battery_discharging_power = 0;
-            int soc = 0;
+            float soc = NAN;
             int soh = 0;
             int charging_cycles = 0;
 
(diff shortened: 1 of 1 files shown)


Sleeper85:
Could you please explain the importance of setting the SoC to NAN when no BMS is combined ?

txubelaxu:
It is because it is a bad meassureament when no data is available. If you
have, for example, an alarm that would rise when SoC arrives to 0%, each
time esp32 restarts/starts would trigger that alarm.
At the same time, the graph of them meassurement SoC(t) is dirty.

I think is more correct N/A when no data is available.

By the way (sorry for asking that), could you delete the other two PR
(closed) that I sent yesterday. I do not know Github enough and is normal I
make mistakes... Sorry about that again. I would apreciate it so much.



Sleeper85 ***@***.***> erabiltzaileak hau idatzi du (2026
ots. 4(a), az. (11:11)):

Sleeper85:
Okay, I understand your approach if you have automations in HA based on the SOC.

Sleeper85:
Why did you close?

txubelaxu:
Hi @Sleeper85,

I needed to delete some data from my "wrong PR" (it was important for me)
and you told me to remove my "wrong" branch. I did it, but from that branch
I rised all the PRs (wrong and good one)... So, this made the good PR
closed... :-(

I'll try to reopen the good ones again.

Sleeper85:
Okay, no problem, I've remembered what you want to do. I'll add it to my to-do list; I have way too many things going on right now...

PR #194, opened 2026-02-03
Component Fault Registry & Advanced LED Control

MrPabloUK:
This PR introduces a complete overhaul of the system’s health monitoring and LED status logic.

It replaces previous LED control with a **Centralized Fault Registry (Backend)** and implements **Synchronized Diagnostic Blink Codes (Frontend)**. This ensures that multiple BMS modules, inverters, and peripherals report errors in a unified, synchronized manner via LED control.

### Centralized Fault Registry

**Dynamic Storage**: A global 2D-vector (global_instance_faults) that automatically tracks health status for up to 16 categories (API, BMS, canbus, etc.) and unlimited instances (multiple batteries, etc) per category.

**Event-Driven**: Faults are updated via a lightweight global function (global_update_fault).

**HA Integration**: Automatically publishes a raw JSON sensor (system_fault_data) containing the complete system health matrix for easy parsing in Home Assistant via a markdown card.

image

```
type: markdown
content: |-
  | Component | Instance Status |
  | :--- | :--- |
  {%- set raw = states('sensor.yambms_demo_system_fault_data') %}
  {%- if raw not in ['unknown', 'unavailable', 'none'] %}
      {%- set data = raw | from_json %}
      {%- set names = ['API / MQTT', 'BMS', 'CAN->INV', 'RS485->INV', 'Shunt', 'Balancer', 'Network', 'Aux 1', 'Cat 8', 'Cat 9', 'Cat 10', 'Cat 11', 'Cat 12', 'Cat 13', 'Cat 14', 'Cat 15'] %}
      {%- for cat_idx in range(0, 16) %}
        {%- set instances = data[cat_idx] %}
        {%- if instances | count > 0 %}
          {%- set ns = namespace(visual='') %}
          {%- for inst in instances %}
            {#- Accessible Icons: Cross/Check for shape-based recognition -#}
            {%- set icon = '❌' if inst == 1 else '✅' %}
            {%- set ns.visual = ns.visual ~ icon ~ '  ' %}
          {%- endfor %}
    | {{ names[cat_idx] }} | {{ ns.visual }} |
        {%- endif %}
      {%- endfor %}
    {%- else %}
  | System | Initializing... |
    {%- endif %}
```

**Fault Mask**: Maintains a 16-bit integer mask (global_system_fault_mask) for efficient checks by other functions.

---

### Synchronized Diagnostic LED Blink Codes

Replaced the previous logic with a structured and efficient logic supporting multi-LED output:

**Heartbeat**: When healthy, LEDs pulse green for 1 second, off for 2 seconds.

**Diagnostic Codes**: When a fault is deteched, LEDs transmit a specific number of blinks corresponding to the Fault Category (e.g., 2 blinks = BMS Error, 5 blinks= Shunt Error).

**LED synchronisation**: On a multi-LED board, all LEDs share a common time base. The 1 second pulse happens simultaneously on all LEDs, making the board easier to read.

**Fault Spotlight Mode**: When a fault is active, healthy LEDs turn off during the blink phase to reduce visual noise and highlight the faulty component.

**Multi-Fault Rotation**: If one LED monitors multiple components (e.g., BMS + Shunt) and both fail, the LED automatically rotates through the error codes sequentially.

---

### Configuration

Users can now map specific components to specific hardware LEDs using the *_status_led_id substitution in their package configuration.

Example:
```
# Assign BMS 1 to the first led
bms_status_led_id: 'led_1'

# Assign Inverter CANBUS to the second led
canbus_status_led_id: 'led_2'

# Shared LED Strategy (Both report to the same LED)
# The LED will rotate codes if both fail.
bms_status_led_id: 'esp_light'
canbus_status_led_id: 'esp_light'
```

---

### Documentation

Updated YamBMS_behaviour.md.
- Blink reference table (1=API / MQTT, 2=BMS, etc.).
- Advanced features (Spotlight, Rotation).

Updated template YAMLs.

--- 

### Breaking Changes

The main YAML will need to be updated, templates have been altered as needed.\
Note that canbus_id and rs485_id must be a number.

### DIFF
--- a/YamBMS_LP_PVbrain2.yaml
+++ b/YamBMS_LP_PVbrain2.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.01.29
-# Version : 1.6.1
+# Updated : 2026.02.03
+# Version : 1.6.2
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -141,6 +141,11 @@ substitutions:
   # +--------------------------------------+
   bms_update_interval: '3s'    # frequency at which BMS data is refreshed, going below '3s' can cause problems
   # +--------------------------------------+
+  # | Health & LED Assignment Logs         |
+  # +--------------------------------------+
+  enable_led_logging: 'false'     # Set to true to see if an led is assigned to a component correctly, logged every 60 seconds
+  enable_health_logging: 'false'  # Set to true to see if component health for all component instances, logged every 60 seconds
+  # +--------------------------------------+
   # | UART expander WK2168                 |
   # +--------------------------------------+
   uart0_baudrate: '2400'    # uart_id : uart_${uart_hub_bus}_0          WKS           {RJ45-UART0}
@@ -162,7 +167,13 @@ substitutions:
 
 packages:
 
-  device_board: !include packages/board/board_ESP32-S3_DevKitC-1_PVbrain2.yaml
+  device_board: 
+    file: packages/board/board_ESP32-S3_DevKitC-1_PVbrain2.yaml
+
+  #   vars:
+  #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
+  #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
+
 
   # +------------------------------------------------------------------------------+
   # | >>> Board Options
@@ -244,10 +255,10 @@ packages:
   canbus1: !include
     file: packages/yambms/yambms_canbus.yaml # yambms_canbus_web_server.yaml
     vars:
-      canbus_id: 'canbus1'
+      canbus_id: '1' # Must be a number
       canbus_name: 'CANBUS 1'
       canbus_node_id: 'canbus_inverter_1' # CAN bus node to which your inverter is connected
-      light_id: 'esp_light'
+      canbus_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
       # The CANBUS link will be considered down if no response from the inverter (ID 0x305) for 5s
       # The Deye inverter sends the ACK (can_id 0x305) only when it receives the can_id 0x356
       canbus_link_timer: '5s'
@@ -261,13 +272,12 @@ packages:
   # rs485_pylon: !include
   #   file: packages/yambms/yambms_rs485_pylon.yaml # yambms_rs485_pylon_web_server.yaml
   #   vars:
-  #     rs485_id: 'rs485_1'
+  #     rs485_id: '1' # Must be a number
   #     rs485_name: 'RS485 1'
   #     rs485_uart_id: 'uart_esp_3' # UART interface to which your inverter is connected
   #     rs485_uart_baud_rate: '9600'
   #     rs485_update_timeout: '5s'
-  #     light_id: 'esp_light'
-
+  #     rs485_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
 
 
   # +------------------------------------------------------------------------------+
--- a/YamBMS_LP_PVbrain2.yaml
+++ b/YamBMS_LP_PVbrain2.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.02.03
-# Version : 1.6.2
+# Updated : 2026.02.04
+# Version : 1.7.0
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -167,7 +167,7 @@ substitutions:
 
 packages:
 
-  device_board: 
+  device_board: !include
     file: packages/board/board_ESP32-S3_DevKitC-1_PVbrain2.yaml
 
   #   vars:
--- a/YamBMS_Remote_Packages_example.yaml
+++ b/YamBMS_Remote_Packages_example.yaml
@@ -175,9 +175,12 @@ packages:
       # - path: 'packages/board/board_ESP32-C3_ETH01-EVO.yaml'
       # - path: 'packages/board/board_ESP32-C3_FlipC3.yaml'
       # - path: 'packages/board/board_ESP32-S3_AtomS3-Lite.yaml'
+      # - path: 'packages/board/board_ESP32-S3_AtomS3.yaml'
+      # - path: 'packages/board/board_ESP32-S3_AtomS3R.yaml'
       # - path: 'packages/board/board_ESP32-S3_DevKitC-1_LED-38.yaml'
       # - path: 'packages/board/board_ESP32-S3_DevKitC-1_LED-48.yaml'
       # - path: 'packages/board/board_ESP32-S3_LilyGo-T-Connect.yaml'
+      # - path: 'packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml'
       # - path: 'packages/board/board_ESP32-S3_WS-ETH.yaml'
       # - path: 'packages/board/board_ESP32-S3_WS-RS485-CAN.yaml'
       # - path: 'packages/board/board_ESP32-S3_WS-Touch-LCD-4.3.yaml'
@@ -189,22 +192,18 @@ packages:
       #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
       #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
 
-      # - path: 'packages/board/board_ESP32-S3_AtomS3.yaml' # 'packages/board/board_ESP32-S3_AtomS3R.yaml'
+  # +------------------------------------------------------------------------------+
+  # | >>> Board Display
+  # | >>> Uncomment the display that corresponds to your board
+  # | >>> Importing the display also activates PSRAM
+  # +------------------------------------------------------------------------------+
+
+      # - path: 'packages/board/board_options_display_128x128.yaml' # AtomS3 / AtomS3R display
+      # - path: 'packages/board/board_options_display_480x222.yaml' # LilyGo T-Connect Pro display
+      
       #   vars:
       #     display_auto_next_page_interval: '5s'
       #     display_rotation: '0' # 0/90/180/270 rotation angle in degrees
-      #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
-      #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
-
-      # # LilyGo T-Connect Pro - Board
-      # - path: 'packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml'
-      #   vars:
-      #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
-      #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
-      # # LilyGo T-Connect Pro - Display - Importing the display also activates PSRAM
-      # - path: 'packages/board/board_options_display_480x222.yaml'
-      #   vars:
-      #     display_auto_next_page_interval: '5s'
 
 
 
--- a/YamBMS_Remote_Packages_example.yaml
+++ b/YamBMS_Remote_Packages_example.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.02.03
-# Version : 1.6.2
+# Updated : 2026.02.04
+# Version : 1.7.0
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -193,9 +193,14 @@ packages:
       #   vars:
       #     display_auto_next_page_interval: '5s'
       #     display_rotation: '0' # 0/90/180/270 rotation angle in degrees
+      #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
+      #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
 
       # # LilyGo T-Connect Pro - Board
       # - path: 'packages/board/board_ESP32-S3_LilyGo-T-Connect-Pro.yaml'
+      #   vars:
+      #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
+      #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
       # # LilyGo T-Connect Pro - Display - Importing the display also activates PSRAM
       # - path: 'packages/board/board_options_display_480x222.yaml'
       #   vars:
--- a/YamBMS_Remote_Packages_example.yaml
+++ b/YamBMS_Remote_Packages_example.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.01.29
-# Version : 1.6.1
+# Updated : 2026.02.03
+# Version : 1.6.2
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -140,7 +140,11 @@ substitutions:
   # | BMS Settings                         |
   # +--------------------------------------+
   bms_update_interval: '3s'    # frequency at which BMS data is refreshed, going below '3s' can cause problems
-
+  # +--------------------------------------+
+  # | Health & LED Assignment Logs         |
+  # +--------------------------------------+
+  enable_led_logging: 'false'     # Set to true to see if an led is assigned to a component correctly, logged every 60 seconds
+  enable_health_logging: 'false'  # Set to true to see if component health for all component instances, logged every 60 seconds
 
 
 # +--------------------------------------+
@@ -180,6 +184,10 @@ packages:
       # - path: 'packages/board/board_ESP32-S3_WS-Touch-LCD-7.yaml'
       # - path: 'packages/board/board_ESP32-S3_XIAO.yaml'
       # - path: 'packages/board/board_ESP32-S3_YBoard.yaml'
+
+      #   vars:
+      #     server_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
+      #     network_status_led_id: 'virtual_status_light' # led used to show status. Default = esp_light, set specific led or set to 'virtual_status_light' to disable
 
       # - path: 'packages/board/board_ESP32-S3_AtomS3.yaml' # 'packages/board/board_ESP32-S3_AtomS3R.yaml'
       #   vars:
@@ -283,12 +291,12 @@ packages:
     files:
       - path: 'packages/yambms/yambms_canbus.yaml' # yambms_canbus_web_server.yaml
         vars:
-          canbus_id: 'canbus1'
+          canbus_id: '1' # Must be a number
           canbus_name: 'CANBUS 1'
-          canbus_node_id: 'canbus_inverter_1'
-          light_id: 'esp_light' # led used to show status. Default = esp_light, comment out or set to 'none' to disable
-          # The CANBUS link will be considered down if no response from the inverter (0x305 frame) for 5s
-          # The Deye inverter sends the ACK 0x305 only when it receives the 0x356 frame.
+          canbus_node_id: 'canbus_inverter_1' # CAN bus node to which your inverter is connected
+          canbus_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
+          # The CANBUS link will be considered down if no response from the inverter (ID 0x305) for 5s
+          # The Deye inverter sends the ACK (can_id 0x305) only when it receives the can_id 0x356
           canbus_link_timer: '5s'
 
 
@@ -304,13 +312,12 @@ packages:
   #   files:
   #     - path: 'packages/yambms/yambms_rs485_pylon.yaml' # yambms_rs485_pylon_web_server.yaml
   #       vars:
-  #         rs485_id: 'rs485_1'
+  #         rs485_id: '1' # Must be a number
   #         rs485_name: 'RS485 1'
   #         rs485_uart_id: 'uart_esp_3' # UART interface to which your inverter is connected
   #         rs485_uart_baud_rate: '9600'
   #         rs485_update_timeout: '5s'
-  #         light_id: 'esp_light' # led used to show status. Default = esp_light, comment out or set to 'none' to disable
-
+  #         rs485_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
 
 
   # +------------------------------------------------------------------------------+
--- a/documents/README/YamBMS_behavior.md
+++ b/documents/README/YamBMS_behavior.md
@@ -81,16 +81,100 @@ SoC behavior:
 2) Else if `the battery is fully charged` real SoC is sent
 3) Else SoC `98%` is sent
 
-## LED status
 
-This only applies to boards with a led, both RGB or mono colour.
+## LED Status & Diagnostic Codes
 
-| Level | Color | Flash | Status |
-| --- | --- | --- | --- |
-Critical | $${\color{red}Red}$$ | Quad flash | HA or MQTT KO + CANBUS KO
-Warning | $${\color{yellow}Yellow}$$ | Triple flash | HA or MQTT OK + CANBUS KO
-Info | $${\color{blue}Blue}$$ | Double flash | HA or MQTT KO + CANBUS OK
-Healthy | $${\color{green}Green}$$ | Single flash | HA or MQTT OK + CANBUS OK
+The system uses Synchronized Diagnostic Blink Codes. 
+This allows the device to communicate specific fault categories visually without requiring a network connection.
+
+###  1. System Healthy
+If no faults are present, the LED indicates the system is alive.
+
+*   **Pattern:** 🟢 Green Pulse (1s) $\rightarrow$ ⚫ OFF (1s).
+*   **Meaning:** All linked components operational.
+*   **Cycle Time:** ~2.0 Seconds.
+
+### 2. Component Fault
+When a fault occurs, the cycle changes to a strict 3-part message structure.
+
+| Phase | Color | Duration | Description |
+| :--- | :--- | :--- | :--- |
+| **1. Header** | 🔵 **Blue** | 1.0s | **"Attention."** Marks the start of a error report cycle. |
+| **2. Gap** | ⚫ **OFF** | 0.5s | **"Wait."** A short pause to prepare for counting. |
+| **3. Code** | 🔴 **Red** | Variable | **"Data."** A set count of short Red blinks. |
+
+#### Reading the Blink Code
+The number of Red blinks corresponds to the Fault Category (see table below).
+To make counting easier, long codes are broken into groups of 4:
+*   **5 Pips:** `* * * *` ... `*`
+*   **9 Pips:** `* * * *` ... `* * * *` ... `*`
+
+#### Category Reference Table
+
+| Blinks | Category | Description |
+| :--: | :--- | :--- |
+| **1** | **API / MQTT** | API or MQTT communication failure. |
+| **2** | **BMS** | Battery Management System communication failure. |
+| **3** | **INV CAN** | Inverter CAN-BUS communication failure. |
+| **4** | **INV RS485** | Inverter Modbus/RS485 communication failure. |
+| **5** | **SHUNT** | Shunt communication failure. |
+| **6** | **BALANCER** | Balancer communication failure. |
+| **7** | **NETWORK** | No wifi or ethernet connection. |
+| **8-16** | **AUX 1-9** | Reserved for future expansion. |
+
+### Advanced Behaviors
+
+#### Spotlight Mode
+When a fault is active anywhere in the system:
+1.  Fault LEDs: Display the Red error code.
+2.  Healthy LEDs: Turn OFF during the "Code" phase
+This reduces visual noise, highlighting exactly which components requires attention.
+
+#### Multi-Fault Rotation
+If a single LED monitors multiple components (e.g., a LED monitoring both the BMS and the Shunt) and both fail simultaneously:
+*   **Cycle 1:** Displays BMS Code (2 Pips).
+*   **Cycle 2:** Displays Shunt Code (5 Pips).
+*   **Repeat:** The LED rotates through active faults one by one.
+
+#### Monochromatic LEDs
+For boards without RGB LEDs:
+*   **Header:** Solid on.
+*   **Code:** Blinking.
+*   **Gap/Healthy:** off.
+
+#### Multi-LED Boards
+For boards with multiple LEDs, it's possible to assign a specific led to a specific component.
+For example, led_1 for BMS 1, led_2 for canbus. See the section below for more details.
+
+### Configuration & LED Assignment
+
+The system links specific hardware LEDs to software components using Substitutions in your YAML package configuration.
+
+Each component package (BMS, canbus, etc.) includes a specific variable to define which LED should report its status.
+
+#### Example Configuration
+In your main YAML file, under `packages`, you assign the Light ID to the `*_status_led_id` variable.
+In most cases, the default values can be used but some boards such as the [LilyGo T-Connect](documents/README/Board_LilyGo_T-Connect.md) have multi-LED functionality.
+
+```yaml
+packages:
+  bms:
+    url: https://github.com/Sleeper85/esphome-yambms
+    ref: main
+    refresh: 60min
+    files:
+      - path: 'packages/bms/bms_combine_JK_RS485_Modbus_bms_standard.yaml'
+        vars:
+          # Unique ID for this specific instance
+          bms_id: '1' # must be a number
+          bms_name: 'JK-BMS 1'
+          bms_address: '0x01' # BMS 1 DIP switch
+          
+          # 1. LINKING THE LED
+          # Set this to the ID of the light component defined in your ESPHome config.
+          # Example: 'esp_light', 'led_1', etc. The board config will show further details
+          bms_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
+```
 
 ## CAN bus link
 
--- a/documents/README/YamBMS_behavior.md
+++ b/documents/README/YamBMS_behavior.md
@@ -115,8 +115,8 @@ To make counting easier, long codes are broken into groups of 4:
 | :--: | :--- | :--- |
 | **1** | **API / MQTT** | API or MQTT communication failure. |
 | **2** | **BMS** | Battery Management System communication failure. |
-| **3** | **INV CAN** | Inverter CAN-BUS communication failure. |
-| **4** | **INV RS485** | Inverter Modbus/RS485 communication failure. |
+| **3** | **INV CAN** | Inverter CAN bus communication failure. |
+| **4** | **INV RS485** | Inverter RS485 bus communication failure. |
 | **5** | **SHUNT** | Shunt communication failure. |
 | **6** | **BALANCER** | Balancer communication failure. |
 | **7** | **NETWORK** | No wifi or ethernet connection. |
@@ -199,10 +199,10 @@ This `5s` delay can be modified in the configuration when importing the canbus p
   canbus1: !include
     file: packages/yambms/yambms_canbus.yaml
     vars:
-      canbus_id: 'canbus1'
+      canbus_id: '1' # Must be a number
       canbus_name: 'CANBUS 1'
       canbus_node_id: 'canbus_node1'
-      canbus_light_id: 'esp_light'
+      canbus_status_led_id: 'esp_light' # led used to show status. Default = esp_light, comment out to disable
       # The CANBUS link will be considered down if no response from the inverter (ID 0x305) for 5s
       canbus_link_timer: '5s'
 ```
--- a/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
+++ b/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
@@ -1,5 +1,5 @@
-# Updated : 2025.12.23
-# Version : 1.1.2
+# Updated : 2026.02.03
+# Version : 1.1.3
 # GitHub  : https://github.com/syssi/esphome-jk-bms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -23,6 +23,7 @@ substitutions:
 
 packages:
   sub_device: !include balancer_base_sub_device.yaml
+  fault_updater: !include balancer_base_led_assign.yaml
 
 ota:
   on_begin:
@@ -66,9 +67,11 @@ binary_sensor:
       id: balancer${bms_id}_online_status
       device_id: balancer_${bms_id}
       name: "online status"
-      on_state:
-        then:
-          - lambda: id(bms${bms_id}_var_ext_balancer_online) = x;
+    on_state:
+      then:
+        - lambda: |-
+            // Pass the current state to the update function
+            id(global_update_fault)(${CAT_BALANCER}, ${bms_id}, !x);
     balancing:
       id: balancer${bms_id}_equalizing
       device_id: balancer_${bms_id}
--- a/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
+++ b/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
@@ -17,13 +17,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/gpl.html>.
 
+packages:
+  balancer_base: !include balancer_base.yaml
+
 substitutions:
   balancer_model: "JK NEEY"
   balancer_protocol: "BLE"
-
-packages:
-  sub_device: !include balancer_base_sub_device.yaml
-  fault_updater: !include balancer_base_led_assign.yaml
 
 ota:
   on_begin:
--- a/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
+++ b/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
@@ -1,4 +1,4 @@
-# Updated : 2026.02.04
+# Updated : 2026.02.05
 # Version : 1.1.3
 # GitHub  : https://github.com/syssi/esphome-jk-bms
 
@@ -17,12 +17,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/gpl.html>.
 
-packages:
-  balancer_base: !include balancer_base.yaml
-
 substitutions:
   balancer_model: "JK NEEY"
   balancer_protocol: "BLE"
+
+packages:
+  balancer_base: !include balancer_base.yaml
 
 ota:
   on_begin:
--- a/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
+++ b/packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
@@ -1,4 +1,4 @@
-# Updated : 2026.02.03
+# Updated : 2026.02.04
 # Version : 1.1.3
 # GitHub  : https://github.com/syssi/esphome-jk-bms
 
@@ -67,11 +67,12 @@ binary_sensor:
       id: balancer${bms_id}_online_status
       device_id: balancer_${bms_id}
       name: "online status"
-    on_state:
-      then:
-        - lambda: |-
-            // Pass the current state to the update function
-            id(global_update_fault)(${CAT_BALANCER}, ${bms_id}, !x);
+      on_state:
+        then:
+          - lambda: id(bms${bms_id}_var_ext_balancer_online) = x;
+          - lambda: |-
+              // Pass the current state to the update function
+              id(global_update_fault)(${CAT_BALANCER}, ${bms_id}, !x);
     balancing:
       id: balancer${bms_id}_equalizing
       device_id: balancer_${bms_id}
--- a/packages/bms/bms_sensors_DEYE_CAN_module_full.yaml
+++ b/packages/bms/bms_sensors_DEYE_CAN_module_full.yaml
@@ -1,5 +1,5 @@
-# Updated : 2025.12.22
-# Version : 1.1.8
+# Updated : 2026.02.01
+# Version : 1.1.9
 # GitHub  : https://github.com/GHswitt/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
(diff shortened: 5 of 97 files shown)


MrPabloUK:
I'll provide more detail about how this system is set up.

**Category** = BMS, shunt, balancer, etc.
**Component instance** = a unique ID that sits within a category, for example BMS 1, shunt 3, etc. 

On boot, two 2D-vectors are created, stored as globals.
- `global_instance_faults` - this is used to store a boolean state for each instance of a component.
- `global_led_registry` - this is used to store a pointer to the defined light used for each instance of a component.

As the boot process continues, each component instance will register itself within the two globals.
This is done by calling a global function, keeping the lambda as simple as possible within the component instance YAML.

**Registering into the fault registry** (`global_instance_faults`):
A binary sensor (normally online status) will call `global_update_fault` (a function) on any state change.
The function `global_update_fault` takes 3 inputs - category (int), instance (int) and fault status (bool).
It then inserts into the `global_instance_faults` at the appropriate index, storing the fault status.
The function will automatically resize the registry as more component instances are registered.

**Registering into the led registry** (`global_led_registry`):
A binary sensor (normally online status) will call `global_register_led` (a function) on boot.
The function `global_register_led` takes 3 inputs - category (int), instance (int) and assigned led (pointer).
It then inserts into the `global_led_registry` at the appropriate index, storing the pointer to the assigned led - this keeps memory usage low.
The function will automatically resize the registry as more component instances are registered.

The default led per category is defined as substitution, so can be overridden by a var / substitution.
Most have a default that points at a virtual led, making it easy to exclude a component instance from the LED behaviour by commenting the var out.

Category IDs and component instance IDs are referenced via substitutions, making it easier to understand.
For example, register / update the fault registry:
```
id(global_update_fault)(${CAT_BMS}, ${bms_id}, !x);

becomes

id(global_update_fault)(2, 1, 0);
```

And the same for the LED registry:
```
id(global_register_led)(${CAT_BMS}, ${bms_id}, id(${bms_status_led_id}));

becomes

id(global_update_fault)(2, 1, pointer to light);
```

MrPabloUK:
A lot of effort and testing has gone into the led control logic.

LED behaviour is controlled by a single, lambda function running at 50ms intervals (20Hz). This loop coordinates all LEDs referenced in `global_led_registry`, keeping them synchronized regardless of how many faults are active.

The loop calculates a 'cycle position' based on the current system time (millis()). This creates a universal timeline that all elements of the loop reference.

Every 50ms, the lambda performs a complete sweep of the `global_led_registry` vector.
It map LEDs to status: The loop iterates through the `global_led_registry` to group components by their assigned LED. It determines the status of each specific LED based solely on the components assigned to it.
This allows for independent operation. A BMS LED can show an error code while the canbus LED remains green.

If a LED is in a fault state, the loop calculates the required message length based on the specific category (e.g., category 6 requires more time than category 1). The global cycle timer expands to accommodate the longest active fault, as blink count increases with category.

The lambda determines the colour and blink pattern for each LED individually:

**Scenario A - LED is Healthy**
Condition - all component instances assigned to this specific LED are reporting OK.
Pattern - the LED executes the heartbeat pattern (green pulse).
This LED ignores faults occurring elsewhere in the system. It continues to pulse green even if other LEDs are flashing red.


**Scenario B: LED is Faulted**
Condition: One or more component instances assigned to this LED are reporting a fault.
Phase 1: Header (0-1.0s)
The LED turns Blue. This marks the start of the error message.
Phase 2: Gap (1.0-1.5s)
The LED turns off, user prepares to count blinks.
Phase 3: Blink code (>1.5s)
The LED flashes Red blinks corresponding to the fault category (e.g., 2 blinks for BMS).

If a single LED is responsible for multiple faulted components (e.g., BMS and Shunt both failed), the LED will rotate through the error codes, displaying one code per full cycle.

MrPabloUK:
**Single-LED vs. Multi-LED behavior**
The underlying registry 'global_led_registry' is identical for both board types. The behaviour differs solely based on how components are mapped to the available physical LEDs.

**Single-LED - aggregated status**
On boards with a single status light, all component instances are mapped to the same led_id.

The LED aggregates the health of the entire system and does the following:
No faults -  the LED pulses green.
Any fault - if any component fails (for example, the shunt), the LED immediately switches to the fault cycle (blue header -> red blinks). The green pulse is disabled until the fault is cleared.
Multiple faults - the LED rotates through the error codes for every faulty category.

**Multi-LED - granular status**
On boards with dedicated LEDs (LilyGo T-Connect, etc.), components can be mapped to specific led_ids.

For example, `bms_status_led_id` points to `led_1`, while `canbus_status_led_id` points to `led_4`.
Each LED calculates its status independently, focusing only on specific component instances assigned to it.

Partial faults - if the BMS fails but the canbus connection is fine:
`led_1` - flashes the fault Code (blue header -> 2 red blinks).
`led_4` - pulses green.
You can glance at the board and instantly see that the canbus is healthy (green) while the BMS has a fault.

In Multi-LED configurations, the LEDs remain synchronized. The blue header on faulted LEDs will align with the start of the green pulse on the healthy LEDs.
If a long error code stretches the loop duration, the healthy LEDs will increase the gap between pulses to match the max fault blink duration.

MrPabloUK:
Finally, there are two debug logs available for users, just in case they need to confirm component instance status, and if a LED is assigned. These are disabled by default, but can be activated by adjusting the following substitutions:

```
  # +--------------------------------------+
  # | Health & LED Assignment Logs         |
  # +--------------------------------------+
  enable_led_logging: 'false'     # Set to true to see if an led is assigned to a component correctly, logged every 60 seconds
  enable_health_logging: 'false'  # Set to true to see if component health for all component instances, logged every 60 seconds
  ```
  
  Health status log:
  ```
[09:31:44.585][I][health_rpt:127]: ┌──────────────────────────────┐
[09:31:44.587][I][health_rpt:128]: │     SYSTEM HEALTH REPORT     │
[09:31:44.587][I][health_rpt:129]: └──────────────────────────────┘
[09:31:44.587][I][health_rpt:133]: Global Fault Mask: 12 (Binary: 00000000 00001100)
[09:31:44.590][I][health_rpt:172]:    API / MQTT : [ OK ] 
[09:31:44.593][I][health_rpt:172]:    BMS        : [ OK ] [ OK ] [ OK ] 
[09:31:44.594][W][health_rpt:170]: >> CAN->INV   : [FAIL] 
[09:31:44.597][W][health_rpt:170]: >> RS485->INV : [FAIL] 
[09:31:44.600][I][health_rpt:172]:    SHUNT      : [ OK ] 
[09:31:44.605][I][health_rpt:172]:    NETWORK    : [ OK ] 
[09:31:44.609][I][health_rpt:179]: --------------------------------
  ```
  
  LED assignment log:
  ```
[09:31:43.142][I][led_rpt:301]: ┌──────────────────────────────┐
[09:31:43.144][I][led_rpt:302]: │    LED REGISTRY ASSIGNMENT   │
[09:31:43.144][I][led_rpt:303]: └──────────────────────────────┘
[09:31:43.144][I][led_rpt:341]:    API / MQTT : [LINK] 
[09:31:43.145][I][led_rpt:341]:    BMS        : [LINK] [LINK] [LINK] 
[09:31:43.145][I][led_rpt:341]:    CAN->INV   : [LINK] 
[09:31:43.145][I][led_rpt:341]:    RS485->INV : [LINK] 
[09:31:43.147][I][led_rpt:341]:    SHUNT      : [LINK] 
[09:31:43.147][I][led_rpt:341]:    NETWORK    : [LINK] 
[09:31:43.147][I][led_rpt:347]: --------------------------------
  ```

File: packages/bms/bms_sensors_DEYE_CAN_module_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_DEYE_CAN_module_full.yaml:1]: Why replace `${canbus_id}` with `${ext_canbus_id}` ?
All the example files use `${canbus_id} (canbus_id: 'canbus_bms_1')` to import this BMS.
In this case, `${canbus_id}` actually corresponds to `${canbus_node_id}`. The examples should use `${canbus_node_id}` for clarity.

For code consistency, place on_state below name: ​​and the lambda at the end.

File: packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml:1]: Why replace `${canbus_id}` with `${ext_canbus_id}` ?
All the example files use `${canbus_id} (canbus_id: 'canbus_bms_1')` to import this BMS.
In this case, `${canbus_id}` actually corresponds to `${canbus_node_id}`. The examples should use `${canbus_node_id}` for clarity.

For code consistency, place on_state below name: ​​and the lambda at the end.

File: packages/bms/bms_sensors_DEYE_CAN_top_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_DEYE_CAN_top_full.yaml:1]: Why replace `${canbus_id}` with `${ext_canbus_id}` ?
All the example files use `${canbus_id} (canbus_id: 'canbus_bms_1')` to import this BMS.
In this case, `${canbus_id}` actually corresponds to `${canbus_node_id}`. The examples should use `${canbus_node_id}` for clarity.

For code consistency, place on_state below name: ​​and the lambda at the end.

File: packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml:1]: Why replace `${canbus_id}` with `${ext_canbus_id}` ?
All the example files use `${canbus_id} (canbus_id: 'canbus_bms_1')` to import this BMS.
In this case, `${canbus_id}` actually corresponds to `${canbus_node_id}`. The examples should use `${canbus_node_id}` for clarity.

For code consistency, place on_state below name: ​​and the lambda at the end.

File: documents/README/YamBMS_behavior.md
Comment (Sleeper85) [inline documents/README/YamBMS_behavior.md:1]: PYLON RS485 is not Modbus.

Could you please adapt the `Description` column of the `Category Reference Table` with this text:

* Inverter CAN bus communication failure.
* Inverter RS485 bus communication failure.

The example YAML in the `CAN bus link` section must also be adapted.

File: packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
Comment (Sleeper85) [inline packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml:1]: The indentation of your on_state is incorrect; you need to indent it two spaces to the right.
This lambda should not be deleted `- lambda: id(bms${bms_id}_var_ext_balancer_online) = x;` !

File: packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml:1]: The indentation of your `on_state` is incorrect; you need to indent it two spaces to the left because it's currently inside the lambda.

For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_PACE_RS485_bms_full.yaml:1]: The indentation of your `on_state` is incorrect; you need to indent it two spaces to the left because it's currently inside the lambda.

For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/bms/bms_sensors_SEPLOS_V3_RS485_sniffer_bms_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_SEPLOS_V3_RS485_sniffer_bms_full.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/shunt/shunt_sensors_FAKE.yaml
Comment (Sleeper85) [inline packages/shunt/shunt_sensors_FAKE.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/shunt/shunt_sensors_Victron_SmartShunt_BLE.yaml
Comment (Sleeper85) [inline packages/shunt/shunt_sensors_Victron_SmartShunt_BLE.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/shunt/shunt_sensors_Victron_SmartShunt_UART.yaml
Comment (Sleeper85) [inline packages/shunt/shunt_sensors_Victron_SmartShunt_UART.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/shunt/shunt_sensors_Victron_SmartShunt_UART_minimal.yaml
Comment (Sleeper85) [inline packages/shunt/shunt_sensors_Victron_SmartShunt_UART_minimal.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/bms/bms_sensors_FAKE.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_FAKE.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/shunt/shunt_sensors_Junctek_KHF_UART.yaml
Comment (Sleeper85) [inline packages/shunt/shunt_sensors_Junctek_KHF_UART.yaml:1]: For code consistency, place `on_state` below `name:` ​​and the lambda at the end.

File: packages/yambms/yambms_canbus.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_canbus.yaml]: Why did you rename ${canbus_id} ?

**Comment by MrPabloUK** [comment]
@GHswitt, I'd be very interested in your test results as you have a multitude of board types with different component instances attached.

File: packages/yambms/yambms_canbus_web_server.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_canbus_web_server.yaml:1]: Replace `${ext_canbus_id}` with `canbus${canbus_id}`, as is done for BMS and Shunt devices.

Please update the date and version.

File: packages/yambms/yambms_rs485_pylon.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_rs485_pylon.yaml:1]: Adding the substitution `ext_rs485_id:` will cause problems if the `rs485` package is imported multiple times. The `rs485_id` variable is currently only used to add a prefix to the `sensor ID` values ​​to ensure uniqueness when the package is imported multiple times.

Replace `${ext_rs485_id}` with `rs485${rs485_id}`, as is done for BMS and Shunt devices.

File: packages/yambms/yambms_rs485_pylon_web_server.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_rs485_pylon_web_server.yaml:1]: Replace `${ext_rs485_id}` with `rs485${rs485_id}`, as is done for BMS and Shunt devices.
Please update the date and version.

File: packages/yambms/yambms_core.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_core.yaml:1]: Please update the date at the top of the YAML.

File: packages/yambms/yambms_service.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_service.yaml:1]: Please update the date at the top of the YAML.

File: YamBMS_LP_PVbrain2.yaml
Comment (Sleeper85) [inline YamBMS_LP_PVbrain2.yaml:1]: Should `!include` be added after `device_board:` ?

File: YamBMS_Remote_Packages_example.yaml
Comment (Sleeper85) [inline YamBMS_Remote_Packages_example.yaml:1]: I'm making a single comment here that will apply to all the example YAML files.
Can you change the version to 1.7.0, it should not change again afterwards (this corresponds to the YamBMS version).

Why place the new `vars:` block between `YBoard` and `AtomS3` ?
It should be below all the boards, in my opinion.

File: packages/yambms/yambms_canbus.yaml
Comment (Sleeper85) [inline packages/yambms/yambms_canbus.yaml:1]: Adding the substitution `ext_canbus_id:` will cause problems if the `canbus` package is imported multiple times. The `canbus_id` variable is currently only used to add a prefix to the `sensor ID` values ​​to ensure uniqueness when the package is imported multiple times.

Replace `${ext_canbus_id}` with `canbus${canbus_id}`, as is done for BMS and Shunt devices.

**Review by Sleeper85 — CHANGES_REQUESTED** [review]

File: packages/bms/bms_sensors_DEYE_CAN_module_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_module_full.yaml:1]: Thank you for spotting this, I made a find and replace across multiple files and mistakenly included BMS files.
I'll revert this change.

File: packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml:1]: Thank you for spotting this, I made a find and replace across multiple files and mistakenly included BMS files.
I'll revert this change.

File: packages/bms/bms_sensors_DEYE_CAN_top_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_top_full.yaml:1]: Thank you for spotting this, I made a find and replace across multiple files and mistakenly included BMS files.
I'll revert this change.

File: packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml:1]: Thank you for spotting this, I made a find and replace across multiple files and mistakenly included BMS files.
I'll revert this change.

File: packages/yambms/yambms_canbus.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus.yaml:1]: The 'ext_canbus_id' substitution does not cause issues during compilation and flash.

Image

However, I'll change to the 'canbus${canbus_id}' pattern.

File: packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
Comment (MrPabloUK) [reply packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_DEYE_CAN_module_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_module_full.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_module_minimal.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_DEYE_CAN_top_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_top_full.yaml:1]: Updated, will push a commit shortly, although there is no on_state code added to this file.

File: packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_DEYE_CAN_top_minimal.yaml:1]: Updated, will push a commit shortly, although there is no on_state code added to this file.

File: packages/bms/bms_sensors_FAKE.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_FAKE.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_JK_UART_4G-GPS_full.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_PACE_RS485_bms_full.yaml:1]: Updated, will push a commit shortly

File: packages/bms/bms_sensors_SEPLOS_V3_RS485_sniffer_bms_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_SEPLOS_V3_RS485_sniffer_bms_full.yaml:1]: Updated, will push a commit shortly

File: packages/shunt/shunt_sensors_FAKE.yaml
Comment (MrPabloUK) [reply packages/shunt/shunt_sensors_FAKE.yaml:1]: Updated, will push a commit shortly

File: packages/shunt/shunt_sensors_Junctek_KHF_UART.yaml
Comment (MrPabloUK) [reply packages/shunt/shunt_sensors_Junctek_KHF_UART.yaml:1]: Updated, will push a commit shortly

File: packages/shunt/shunt_sensors_Victron_SmartShunt_BLE.yaml
Comment (MrPabloUK) [reply packages/shunt/shunt_sensors_Victron_SmartShunt_BLE.yaml:1]: Updated, will push a commit shortly

File: packages/shunt/shunt_sensors_Victron_SmartShunt_UART.yaml
Comment (MrPabloUK) [reply packages/shunt/shunt_sensors_Victron_SmartShunt_UART.yaml:1]: Updated, will push a commit shortly

File: packages/shunt/shunt_sensors_Victron_SmartShunt_UART_minimal.yaml
Comment (MrPabloUK) [reply packages/shunt/shunt_sensors_Victron_SmartShunt_UART_minimal.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_canbus.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_canbus_web_server.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus_web_server.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_core.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_core.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_rs485_pylon.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_rs485_pylon.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_rs485_pylon_web_server.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_rs485_pylon_web_server.yaml:1]: Updated, will push a commit shortly

File: YamBMS_LP_PVbrain2.yaml
Comment (MrPabloUK) [reply YamBMS_LP_PVbrain2.yaml:1]: Updated, will push a commit shortly

File: packages/yambms/yambms_service.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_service.yaml:1]: Updated, will push a commit shortly

File: YamBMS_Remote_Packages_example.yaml
Comment (MrPabloUK) [reply YamBMS_Remote_Packages_example.yaml:1]: I'll set all example files to 1.7.0.

Regarding the vars, we have two approaches.

Include in var block for each 'type' of board.
In this case, the new var block stays where it is, but the *_status_led_id vars are also added to `ESP32-S3_AtomS3` and `ESP32-S3_LilyGo-T-Connect-Pro`.
This is my preferred approach as it's easier for a user to comment appropriately.

Place in var block below all boards.
In this case, a user of the two above boards will need to delete or leave the var header commented out.

For now, I'll use approach 1 but we can discuss further.

File: documents/README/YamBMS_behavior.md
Comment (MrPabloUK) [reply documents/README/YamBMS_behavior.md:1]: Updated, will push a commit shortly

File: packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
Comment (Sleeper85) [inline packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml:1]: Why did you call the package `fault_updater:` when importing it, instead of `led_assign:` elsewhere ?

Could you create a `balancer_base.yaml` file to group `balancer_base_sub_device.yaml` and `balancer_base_led_assign.yaml` as we do for BMS and Shunt + update the other YAML files to use `balancer_base.yaml`.

File: packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
Comment (Sleeper85) [inline packages/bms/bms_sensors_PACE_RS485_bms_full.yaml:1]: Please delete the empty line 91.

File: packages/yambms/yambms_canbus.yaml
Comment (Sleeper85) [reply packages/yambms/yambms_canbus.yaml:1]: > The 'ext_canbus_id' substitution does not cause issues during compilation and flash.

That's true for compilation and flash, but it's surely a bug if the `canbus` package is imported twice because the IDs of all canbus entities will have the same prefix.

File: packages/yambms/yambms_canbus_web_server.yaml
Comment (Sleeper85) [reply packages/yambms/yambms_canbus_web_server.yaml:1]: Replace `${canbus_id}` with `canbus_${canbus_id}`.

File: packages/yambms/yambms_rs485_pylon.yaml
Comment (Sleeper85) [reply packages/yambms/yambms_rs485_pylon.yaml:1]: Replace the ids `rs485_rs485${rs485_id}` with `rs485_${rs485_id}`.

File: packages/yambms/yambms_canbus.yaml
Comment (Sleeper85) [reply packages/yambms/yambms_canbus.yaml:1]: Replace the ids `canbus_canbus${canbus_id}` with `canbus_${canbus_id}`.

File: YamBMS_Remote_Packages_example.yaml
Comment (Sleeper85) [reply YamBMS_Remote_Packages_example.yaml:1]: I've found this section a bit messy for the display boards for a while now. I've just pushed a commit to split them in two; I think that will be clearer. Could you do the same for all the example YAML files ?

I'm taking care of adapting the AtomS3 and S3R boards accordingly.

Many thanks

**Review by Sleeper85 — CHANGES_REQUESTED** [review]

File: packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
Comment (MrPabloUK) [reply packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml:1]: Updated

File: packages/bms/bms_sensors_PACE_RS485_bms_full.yaml
Comment (MrPabloUK) [reply packages/bms/bms_sensors_PACE_RS485_bms_full.yaml:1]: Updated

File: packages/yambms/yambms_canbus.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus.yaml:1]: Updated

File: packages/yambms/yambms_rs485_pylon.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_rs485_pylon.yaml:1]: Updated

File: packages/yambms/yambms_canbus_web_server.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus_web_server.yaml:1]: This error is not present in this file.

File: packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml
Comment (MrPabloUK) [reply packages/balancer/balancer_sensors_JK_NEEY_BLE.yaml:1]: The package was originally handling additional functions, the name was changed but this package alias was not updated.

File: packages/yambms/yambms_canbus.yaml
Comment (MrPabloUK) [reply packages/yambms/yambms_canbus.yaml:1]: Regarding your point about `ext_canbus_id`, the substitution will be changed on each package import.
This is because this substitution is derived from another substitution / var.

Package 1:
`canbus_id: 1` results in `ext_canbus_id: "canbus${canbus_id}"` being replaced with `ext_canbus_id: "canbus1"`

Package 2:
`canbus_id: 2` results in `ext_canbus_id: "canbus${canbus_id}"` being replaced with `ext_canbus_id: "canbus2"`

This is why you see the following, not `switch.canbus_1_inverter_heartbeat_monitoring2`.
image

This is no longer a part of this commit, but cascading substitutions can be used if required.

File: YamBMS_Remote_Packages_example.yaml
Comment (MrPabloUK) [reply YamBMS_Remote_Packages_example.yaml:1]: A much cleaner setup, changes made to example YAML files.

**Comment by MrPabloUK** [comment]
@Sleeper85, I appreciate the file level review comments. It's making it much easier to make necessary changes whilst stuck in a very long work call.

**Comment by GHswitt** [comment]
@MrPabloUK Great work, I love this PR! I will test it over the next days :-)

**Comment by GHswitt** [comment]
@MrPabloUK: I'm testing it on my balancer node, the code looks like this:

```
  # Balancer for BMS 3: Basen 1
  balancer3: !include
    file: packages/balancer/balancer_modbus_JK_NEEY_BLE.yaml
    vars:
      bms_id: '3' # must be a number
      balancer_name: 'NEEY 1'
      balancer_ble_mac_address: !secret neey1_mac
      balancer_status_led_id: 'esp_light'

  # Balancer for BMS 4: Basen 2
  balancer4: !include
    file: packages/balancer/balancer_modbus_JK_NEEY_BLE.yaml
    vars:
      bms_id: '4' # must be a number
      balancer_name: 'NEEY 2'
      balancer_ble_mac_address: !secret neey2_mac
      balancer_status_led_id: 'esp_light'
```

It only controls two balancers. These balancers are connected to battery/BMS 3 and 4. So there are two instances of balancers.
When I disable bluetooth to the first balancer (balancer3), I get this view:

image

With all enabled:

image

If I disable the second balancer only (balancer4) I get:

image

With both disabled:

image

Both enabled again:

image

So I'm not sure if this is the expected behavior. I would either expect only two instances, which I understand is difficult because of the `bms_ids` and compile time only substitutions or see all `four` from the start.

**Comment by GHswitt** [comment]
@MrPabloUK There is a category for RS485 to the inverter, would it be possible to add a RS485 category for the multi-node communication? Like `CAT_NODE_RS485`?

**Comment by MrPabloUK** [comment]
That's interesting, @GHswitt. It suggests that your balancers are not registering correctly. I'll confess that I'm not that familiar with how the standalone balancers integrate, so I'll need to do some investigation.
On initial boot, do you see 2 balancers in the fault matrix?
Edit:
On second thought, are these the only standalone balancers? If so, the first id is 3. I didn't test the lowest id number being >1, so I'll examine that tonight.

As for your question about multi node fault tracking, would this be on the sub nodes only, or all nodes?

**Comment by MrPabloUK** [comment]
@GHswitt, I've updated the fault registry to handle unassigned IDs.
When registering a component id, the missing indexes were created but with a default value of false for the error.
That worked fine, as long as eventually the real component instance appeared.

In your case, that would never happen.

Now, the component instance is mapped into one of three states:
0 = Healthy
1 = Fault
2 = Unassigned

This should resolve the issue.
Update your markdown card as follows:

```
type: markdown
content: |-
  | Component | Instance Status |
  | :--- | :--- |
  {%- set raw = states('sensor.yambms_demo_system_fault_data') %}
  {%- if raw not in ['unknown', 'unavailable', 'none'] %}
      {%- set data = raw | from_json %}
      {%- set names = ['API / MQTT', 'BMS', 'CAN->INV', 'RS485->INV', 'Shunt', 'Balancer', 'Network', 'Aux 1', 'Cat 8', 'Cat 9', 'Cat 10', 'Cat 11', 'Cat 12', 'Cat 13', 'Cat 14', 'Cat 15'] %}
      {%- for cat_idx in range(0, 16) %}
        {%- set instances = data[cat_idx] %}
        {%- if instances | count > 0 %}
          {%- set ns = namespace(visual='') %}
          {%- for inst in instances %}
            {#- Determine instance state -#}
            {%- if inst == 1 %}
               {#- Fault -#}
               {%- set icon = '❌' %} 
            {%- elif inst == 0 %}
               {#- Healthy -#}
               {%- set icon = '✅' %} 
            {%- else %}
               {#- Unregistered -#}
               {%- set icon = ' ' %} 
            {%- endif %}
            {#- Use Em Space ( ) as separator -#}
            {%- set ns.visual = ns.visual ~ icon ~ '  ' %}
          {%- endfor %}
    | {{ names[cat_idx] }} | {{ ns.visual }} |
        {%- endif %}
      {%- endfor %}
    {%- else %}
  | System | Initializing... |
    {%- endif %}
```
    
This is exactly why your setup is a near perfect testbed! It makes my 2 BMS, 1 inverter configuration look very boring.

**Comment by GHswitt** [comment]
Yes, this node handles only the standalone balancers. I will try the updated code.

I currently only have the problem that after updating my YamBMS2 node (6x BMS), the main node has lost CAN status and the inverter does not see YamBMS anymore.

I don't know why, I did not update the main node and it looks fine. But the CAN lost time matches the update of the BMS node:

image

image

@Sleeper85 
Edit: Ok, I got it working again after enabling "Discharge Enabled".

So to summarize:

1. Set "Discharge Enabled" to False (I only want to charge the batteries)
2. Update the BMS node (BMS will go offline and online again)
3. CAN is offline, inverter does not see anything. But YamBMS status was fine, all BMS available and visible.
4. Set "Discharge Enabled" to True
5. CAN is online again

This behaviour is different from before. I updated the main node two days ago. And the last month I had often "Discharge Enabled" to False. CAN status normally was always online.

**Comment by MrPabloUK** [comment]
In order to get CAN bus to Inverter status to not stay as unknown on boot, I added code to force it as offline on start up. Otherwise, it was possible to stay in an 'unknown' state.

```
esphome:
  on_boot:
    - priority: 600.0
      then:
        - lambda: |-
            // Register System
            id(global_register_led)(${CAT_INV_CAN_BUS}, ${canbus_id}, id(${canbus_status_led_id}));
            // Force register fault as default, will be updated once CANBUS is working
            // (true = fault)
            id(global_update_fault)(${CAT_INV_CAN_BUS}, ${canbus_id}, true);          
    - priority: -100.0
      then:
        - lambda: |-
            // Initialise canbus status as off
            id(canbus${canbus_id}_status).publish_state(false);
```

I didn't consider this would mean no can packet would be sent to the inverter straight away. Instead, it would wait for 30 seconds as per the following code:

```
interval:
  - interval: 30s
    then:
      - lambda: |-
          // Test inverter 0x305 ACK
          if (id(canbus${canbus_id}_send_canbus_data) == false && id(${yambms_id}_var_initialized) == true)
          {
            id(canbus${canbus_id}_send_canbus_data) = true;                      // Sending CANBUS frames allowed
            id(canbus${canbus_id}_script_canbus_link_timer).execute();           // Restart CANBUS link timer
            ESP_LOGD("yambms_debug", "CAN bus starting...");
          }
```

I'll review this now.

**Comment by GHswitt** [comment]
Another note: packages\board\board_ESP32-S3_LilyGo-T-Connect-Pro.yaml does not compile anymore. It does not have a LED:

```
Couldn't find ID 'global_register_led'. Please check you have defined an ID with that name in your configuration. These IDs look similar: "global_instance_faults".
```

Adding `led_control: !include board_options_led_control.yaml` alone does not help as it does not have an `esp_light`:

```
Couldn't find ID 'esp_light'. Please check you have defined an ID with that name in your configuration. These IDs look similar: "back_light".
```

**Comment by MrPabloUK** [comment]
@GHswitt, what is the behaviour of the affected node when using code from the current dev branch?~~

Re. the t-connect pro board, the board config has no led configured. I'll look at the datasheet and see what the led pin is.

**Comment by GHswitt** [comment]
> In order to get CAN bus to Inverter status to not stay as unknown on boot, I added code to force it as offline on start up. Otherwise, it was possible to stay in an 'unknown' state.
> 
> ```
> esphome:
>   on_boot:
>     - priority: 600.0
>       then:
>         - lambda: |-
>             // Register System
>             id(global_register_led)(${CAT_INV_CAN_BUS}, ${canbus_id}, id(${canbus_status_led_id}));
>             // Force register fault as default, will be updated once CANBUS is working
>             // (true = fault)
>             id(global_update_fault)(${CAT_INV_CAN_BUS}, ${canbus_id}, true);          
>     - priority: -100.0
>       then:
>         - lambda: |-
>             // Initialise canbus status as off
>             id(canbus${canbus_id}_status).publish_state(false);
> ```
> 
> I didn't consider this would mean no can packet would be sent to the inverter straight away. Instead, it would wait for 30 seconds as per the following code:
> 
> ```
> interval:
>   - interval: 30s
>     then:
>       - lambda: |-
>           // Test inverter 0x305 ACK
>           if (id(canbus${canbus_id}_send_canbus_data) == false && id(${yambms_id}_var_initialized) == true)
>           {
>             id(canbus${canbus_id}_send_canbus_data) = true;                      // Sending CANBUS frames allowed
>             id(canbus${canbus_id}_script_canbus_link_timer).execute();           // Restart CANBUS link timer
>             ESP_LOGD("yambms_debug", "CAN bus starting...");
>           }
> ```
> 
> I'll review this now.

I have not updated the main node with the CAN<->inverter connection. I've updated the BMS node and after its restart the main node disabled CAN. The main node did not contain this PR but dev as of two days ago.

**Comment by GHswitt** [comment]
> @GHswitt, what is the behaviour of the affected node when using code from the current dev branch?~~
> 
> Re. the t-connect pro board, the board config has no led configured. I'll look at the datasheet and see what the led pin is.

The T-Connect Pro does not have a (standalone) LED. It has a signal labeled ESP_LED in the circuit diagram (GPIO8) but do _not_ use it. It is used for controlling the relay, the labeling is not correct.

image

**Comment by MrPabloUK** [comment]
@Sleeper85, can you update the T-Connect pro config to have a virtual LED, please?

**Comment by GHswitt** [comment]
Unfortunately the balancer node does not work correctly anymore.
I saw on the main node that the balancer node has massive Modbus timeouts:
```
[23:05:44.054][W][modbus:040]: Stop waiting for response from 104 2009ms after last send
[23:05:49.940][W][modbus:040]: Stop waiting for response from 103 2006ms after last send
[23:05:51.948][W][modbus:040]: Stop waiting for response from 104 2007ms after last send
```

When checking the log of the balancer node I see a very high rate of debug messages:
```
[23:09:52.672][D][light:091]: 'esp_light' Setting:
[23:09:52.672][D][light:079]:   Brightness: 50%
[23:09:52.672][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.698][D][light:091]: 'esp_light' Setting:
[23:09:52.698][D][light:079]:   Brightness: 50%
[23:09:52.699][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.755][D][light:091]: 'esp_light' Setting:
[23:09:52.760][D][light:079]:   Brightness: 50%
[23:09:52.760][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.803][D][light:091]: 'esp_light' Setting:
[23:09:52.806][D][light:079]:   Brightness: 50%
[23:09:52.806][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.857][D][light:091]: 'esp_light' Setting:
[23:09:52.861][D][light:079]:   Brightness: 50%
[23:09:52.861][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.912][D][light:091]: 'esp_light' Setting:
[23:09:52.920][D][light:079]:   Brightness: 50%
[23:09:52.920][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:52.966][D][light:091]: 'esp_light' Setting:
[23:09:52.969][D][light:079]:   Brightness: 50%
[23:09:52.969][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.020][D][light:091]: 'esp_light' Setting:
[23:09:53.023][D][light:079]:   Brightness: 50%
[23:09:53.023][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.074][D][light:091]: 'esp_light' Setting:
[23:09:53.080][D][light:079]:   Brightness: 50%
[23:09:53.080][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.128][D][light:091]: 'esp_light' Setting:
[23:09:53.131][D][light:079]:   Brightness: 50%
[23:09:53.131][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.181][D][light:091]: 'esp_light' Setting:
[23:09:53.185][D][light:079]:   Brightness: 50%
[23:09:53.185][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.239][D][light:091]: 'esp_light' Setting:
[23:09:53.245][D][light:079]:   Brightness: 50%
[23:09:53.245][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.291][D][light:091]: 'esp_light' Setting:
[23:09:53.294][D][light:079]:   Brightness: 50%
[23:09:53.294][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.348][D][light:091]: 'esp_light' Setting:
[23:09:53.358][D][light:079]:   Brightness: 50%
[23:09:53.358][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.410][D][light:091]: 'esp_light' Setting:
[23:09:53.412][D][light:079]:   Brightness: 50%
[23:09:53.412][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.457][D][light:091]: 'esp_light' Setting:
[23:09:53.460][D][light:079]:   Brightness: 50%
[23:09:53.460][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:53.510][D][light:091]: 'esp_light' Setting:
[23:09:53.562][D][light:104]:   State: OFF
[23:09:53.562][D][light:091]: 'esp_light' Setting:
[23:09:53.615][D][light:091]: 'esp_light' Setting:
[23:09:53.675][D][light:091]: 'esp_light' Setting:
[23:09:53.722][D][light:091]: 'esp_light' Setting:
[23:09:53.778][D][light:091]: 'esp_light' Setting:
[23:09:53.885][D][light:091]: 'esp_light' Setting:
[23:09:53.885][D][light:091]: 'esp_light' Setting:
[23:09:53.939][D][light:091]: 'esp_light' Setting:
[23:09:54.005][D][light:091]: 'esp_light' Setting:
[23:09:54.049][D][light:091]: 'esp_light' Setting:
[23:09:54.100][D][light:091]: 'esp_light' Setting:
[23:09:54.161][D][light:091]: 'esp_light' Setting:
[23:09:54.201][D][light:091]: 'esp_light' Setting:
[23:09:54.251][D][light:091]: 'esp_light' Setting:
[23:09:54.310][D][light:091]: 'esp_light' Setting:
[23:09:54.355][D][light:091]: 'esp_light' Setting:
[23:09:54.417][D][light:091]: 'esp_light' Setting:
[23:09:54.488][D][light:091]: 'esp_light' Setting:
[23:09:54.514][D][light:091]: 'esp_light' Setting:
[23:09:54.566][D][light:091]: 'esp_light' Setting:
[23:09:54.623][D][light:091]: 'esp_light' Setting:
[23:09:54.671][D][light:091]: 'esp_light' Setting:
[23:09:54.724][D][light:091]: 'esp_light' Setting:
[23:09:54.791][D][light:091]: 'esp_light' Setting:
[23:09:54.827][D][light:091]: 'esp_light' Setting:
[23:09:54.880][D][light:091]: 'esp_light' Setting:
[23:09:55.015][D][light:091]: 'esp_light' Setting:
[23:09:55.015][D][light:091]: 'esp_light' Setting:
[23:09:55.065][D][light:091]: 'esp_light' Setting:
[23:09:55.127][D][light:091]: 'esp_light' Setting:
[23:09:55.152][D][light:091]: 'esp_light' Setting:
[23:09:55.200][D][light:091]: 'esp_light' Setting:
[23:09:55.248][D][light:091]: 'esp_light' Setting:
[23:09:55.303][D][light:091]: 'esp_light' Setting:
[23:09:55.418][D][light:091]: 'esp_light' Setting:
[23:09:55.418][D][light:091]: 'esp_light' Setting:
[23:09:55.462][D][light:091]: 'esp_light' Setting:
[23:09:55.514][D][light:091]: 'esp_light' Setting:
[23:09:55.570][D][light:091]: 'esp_light' Setting:
[23:09:55.623][D][light:091]: 'esp_light' Setting:
[23:09:55.681][D][light:091]: 'esp_light' Setting:
[23:09:55.735][D][light:091]: 'esp_light' Setting:
[23:09:55.776][D][light:091]: 'esp_light' Setting:
[23:09:55.873][D][light:091]: 'esp_light' Setting:
[23:09:55.935][D][light:091]: 'esp_light' Setting:
[23:09:55.935][D][light:091]: 'esp_light' Setting:
[23:09:55.993][D][light:091]: 'esp_light' Setting:
[23:09:56.060][D][light:091]: 'esp_light' Setting:
[23:09:56.093][D][light:091]: 'esp_light' Setting:
[23:09:56.146][D][light:091]: 'esp_light' Setting:
[23:09:56.199][D][light:091]: 'esp_light' Setting:
[23:09:56.254][D][light:091]: 'esp_light' Setting:
[23:09:56.307][D][light:091]: 'esp_light' Setting:
[23:09:56.368][D][light:091]: 'esp_light' Setting:
[23:09:56.427][D][light:091]: 'esp_light' Setting:
[23:09:56.476][D][light:091]: 'esp_light' Setting:
[23:09:56.536][D][light:091]: 'esp_light' Setting:
[23:09:56.567][D][light:091]: 'esp_light' Setting:
[23:09:56.619][D][light:091]: 'esp_light' Setting:
[23:09:56.677][D][light:091]: 'esp_light' Setting:
[23:09:56.733][D][light:091]: 'esp_light' Setting:
[23:09:56.787][D][light:091]: 'esp_light' Setting:
[23:09:56.836][D][light:091]: 'esp_light' Setting:
[23:09:56.886][D][light:091]: 'esp_light' Setting:
[23:09:56.943][D][light:091]: 'esp_light' Setting:
[23:09:57.048][D][light:091]: 'esp_light' Setting:
[23:09:57.049][D][light:091]: 'esp_light' Setting:
[23:09:57.103][D][light:091]: 'esp_light' Setting:
[23:09:57.157][D][light:091]: 'esp_light' Setting:
[23:09:57.209][D][light:091]: 'esp_light' Setting:
[23:09:57.272][D][light:091]: 'esp_light' Setting:
[23:09:57.319][D][light:091]: 'esp_light' Setting:
[23:09:57.319][D][light:104]:   State: ON
[23:09:57.325][D][light:079]:   Brightness: 50%
[23:09:57.325][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.380][D][light:091]: 'esp_light' Setting:
[23:09:57.429][D][light:079]:   Brightness: 50%
[23:09:57.429][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.429][D][light:091]: 'esp_light' Setting:
[23:09:57.435][D][light:079]:   Brightness: 50%
[23:09:57.435][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.482][D][light:091]: 'esp_light' Setting:
[23:09:57.488][D][light:079]:   Brightness: 50%
[23:09:57.488][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.543][D][light:091]: 'esp_light' Setting:
[23:09:57.590][D][light:079]:   Brightness: 50%
[23:09:57.590][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.592][D][light:091]: 'esp_light' Setting:
[23:09:57.597][D][light:079]:   Brightness: 50%
[23:09:57.598][D][light:115]:   Red: 0%, Green: 100%, Blue: 0%
[23:09:57.628][D][text_sensor:119]: 'esp32_date_time_now' >> '2026-02-04 23:09:57'
[23:09:57.632][D][sensor:129]: 'Debug Heap free percent' >> 20.53 %
[23:09:57.632][W][heltec_balancer_ble:264]: [33:39:04:00:72:98] Not connected
[23:09:57.633][W][heltec_balancer_ble:264]: [38:36:12:00:BE:5B] Not connected
```

Also BLE is not working reliable anymore, it disconnects/connects often.
Edit: Switching back fixed the problems.

**Comment by MrPabloUK** [comment]
Could you provide the YAML for your balancer node?
What is the behaviour on the balancer node if logging level is set to WARN?

**Comment by GHswitt** [comment]
[yambms4_Balancer.yaml](https://github.com/user-attachments/files/25083116/yambms4_Balancer.yaml)

It seems the high log output comes from d44f3921ddb921c09db8e5555ec99507b9fff5f4 which replaces the old 1s LED interval with a 50ms interval handling the LEDs.

I've to try with WARN. But even without reading the logs, it creates timeouts on the Modbus. When I try to read the logs it goes nearly completely offline as it struggles to transfer the log output.

**Comment by MrPabloUK** [comment]
@GHswitt, the led control loop has been updated to prevent spamming the log with unnecessary light updates.
It now checks the LED state, rather than just forcing an update every 50ms.
Please let me know what the performance impact is.

File: packages/yambms/yambms_canbus.yaml
Comment (Sleeper85) [reply packages/yambms/yambms_canbus.yaml:1]: The HA ID entity is derived from `name:` ​​not from `id:`.

If you create a substitution with the same name `ext_bms_id` in multiple places, it will be overwritten and return a single value and non-unique `id`.

To be sure, I tested it:

```YAML
substitutions:
  ext_bms_id: bms${bms_id}
```

And I'm getting dozens of error lines regarding the IDs because they all have `bms3`!

```cpp
Couldn't find ID 'bms2_max_temperature_sensor'. Please check you have defined an ID with that name in your configuration. These IDs look similar: "bms3_max_temperature_sensor", "yambms1_max_temperature_sensor", "bms3_min_temperature_sensor".
```

Sleeper85 (approved):
I've fixed the last errors, it's OK now.

GHswitt:
The RP2040 board does not compile after adding the fault registry:

```
packages\board\..\base\device_base_fault_registry.yaml: In lambda function:
packages\board\..\base\device_base_fault_registry.yaml:133:38: error: 'bitset' is not a member of 'std'
packages\board\..\base\device_base_fault_registry.yaml:124:1: note: 'std::bitset' is defined in header '<bitset>'; this is probably fixable by adding '#include <bitset>'
packages\board\..\base\device_base_fault_registry.yaml:133:55: error: request for member 'to_string' in 'mask', which is of non-class type 'uint16_t' {aka 'short unsigned int'}
```

GHswitt:
This is the output of three of my nodes (one has no Wifi/Ethernet):

image

YamBMS: Main node, CAN inverter, 6xBMS via RS485, 5xShunt via RS485, 2xBalancer via RS485
YamBMS2: BMS node, 6xBMS
YamBMS4: Balancer node, 2xNEEY balancer

MrPabloUK:
I'd like to create an issue with this latest feedback, if acceptable?
I need to understand why can bus to inverter is not registered, nor the shunts on node 4.
Again, I appreciate your support in testing this, @GHswitt.

Sleeper85:
I forgot to tell you, congratulations for this beautiful PR 😍

PR #200, opened 2026-02-05
Add M5Stack Tab5

RAR:
This adds the ESP32-P4 based Tab5.  I'm running this as my main yambms node.  I had previously shown off a Tab5 based display - this integrates that display into the main yambms node.   The P4 is pretty powerful compared to the S3 and has a ton more PSRAM.

I had some odd requirements so I'm running a commu module with it to gain a CAN and an RS485.  The Tab also has a "Port A" on the device itself along with an rs485 port.  

display

(please excuse my wiring/fingerprint mess :rofl:)


.gitignore
Sleeper85, line 1: Please exclude this file; it should not be part of your PR.

Sleeper85 (changes requested):

Sleeper85:
Hi,

**Important:** There were significant changes to the `dev` branch this morning.
Following the PR #194 there will be changes coming to boards without integrated LEDs.
I'll keep you informed when it's finalized so you can do your testing.

I also changed the way we import a display this morning.
We'll all look at that together...

RAR:
Ohhh fancy!  I look forward to it.

MrPabloUK:
@RAR, would you be able to provide some more pictures of this board, including with the commu module?
I have a user that might be quite interested in this.

RAR:
> @RAR, would you be able to provide some more pictures of this board, including with the commu module? I have a user that might be quite interested in this.

I'll take some of the whole rig the next time I pull it out.

RAR:
image

image

image

Here ya go!

Sleeper85:
The two commits above allow you to compile the ESP32-P4 Tab5 board even without the LVGL code.
Remember to always make your YAML files independent.
I'll get back to you about this soon.

Sleeper85:
I would also like the LVGL code to be separate from the lambdas that update the LVGL display.
This would allow for lambdas based on ESPHome YamBMS sensors or directly from Home Assistant sensors for standalone displays.

Here's how to name your YAMLs :

Also consider that your 1280x720 design could also be used for another display of the same size.
This means that the term Tab5 could be replaced with something else (RAR maybe ?) so that it's not solely associated with the Tab5.

* board_options_display_lvgl_1280x720_Tab5.yaml (Your LVGL design)
* board_options_display_lvgl_1280x720_Tab5_esphome_sensors.yaml (ESPHome lambdas version)
* board_options_display_lvgl_1280x720_Tab5_ha_sensors.yaml (HA lambdas version)
* board_options_display_lvgl_1280x720_Tab5_esphome.yaml (The YAML to import for use on the YamBMS node)
* board_options_display_lvgl_1280x720_Tab5_ha.yaml (The YAML to import for use on the standalone display)

### Could you also check if it would be possible to retrieve my `LVGL bindings` for the update of your LVGL design ?

If necessary, I could split the lambdas into two (label and obj) to make them reusable ?
The colors are affected via substitutions declared in the design.

* board_options_display_lvgl_esphome_sensors.yaml
* board_options_display_lvgl_ha_sensors.yaml

The images you're using for the battery are high-resolution, which prevented me from flashing your LVGL code onto the WS ESP32-S3 Touch LCD 7" (I had to reduce their size to make it work). You could also do without them and design a 100% LVGL battery like I did, which is lighter ?

You can develop your display in a virtual environment without having to flash your ESP32-P4; see the [SDL2 Display on host platform](https://esphome.io/components/display/sdl/).

Consider using the `yambms_system_fault_data` sensor to display the status of different instances instead of the historical sensors. You will have access to much more information.

It's great to have dedicated displays for YamBMS, thanks for your PR 😉

[You will find more information about the display that I posted yesterday in the forum.](https://diysolarforum.com/threads/yambms-jk-bms-can-with-new-cut-off-charging-logic-open-source.79325/page-219#post-1742758)

RAR:
I'm traveling this weekend, but can take deeper look Monday.

RAR:
Ditched all images and stole your LVGL based battery image.  Split up the files as suggested.   Let me know if this works.  :beers:

Sleeper85:
Good work, what do you think about displaying the **delta_cell_v** in `mV` ?

The current and power units are lowercase; it would be better that way => `A` + `kW`

Regarding the status, keep in mind that the inverter output can be CAN or RS485.

I like your navigation bar; I think I'll take inspiration from it 😉

image

Sleeper85:
Please check the points in my last comment.

Sleeper85:
Please add your esphome version to the example YAMLs and your HA version to the YAML `YamBMS_RP_HA_Display.yaml`.

Issue #182, opened 2026-01-31
Compliance of `Auto` functions for YamBMS 1.7.0


Sleeper85:
@MrPabloUK @GHswitt 

In general, and with a view to converting YamBMS to a native component, I would like these functions to be converted first, with pure lambdas publishing the results to global variables **or/and** to sensors.

For functions that reduce CCL or DCL, it is now necessary to integrate them into the `3-step loop` for calculating the requested current.
There is one loop for CCL and one loop for DCL.

`STEP 2` is where the `Auto` functions perform their calculation based on the maximum current calculated in `STEP 1`.
The publication of the requested values ​​occurs in `STEP 3`.

Replace the lambda in the `sensors` with a lambda in an `interval` (no **return** required) : 

```YAML
esphome:
  on_boot:
    priority: -100.0
    then:
      - lambda: |-
          id(${yambms_id}_var_auto_ccl_functions_count)++; // If your function reduces CCL
          id(${yambms_id}_var_auto_dcl_functions_count)++; // If your function reduces DCL

interval:
  - interval: ${yambms_update_interval}
    then:
      - lambda: |-
          // If your function reduces CCL
          int cc_step = id(${yambms_id}_var_charge_current_step);
          if (cc_step >= 1 && cc_step < id(${yambms_id}_var_auto_ccl_functions_count) + 1)
          {
            // Your C++ code
            // ...
            // ...
            // ...
            id(${yambms_id}_var_charge_current_step)++;
          }

          // If your function reduces DCL
          int dc_step = id(${yambms_id}_var_discharge_current_step);
          if (dc_step >= 1 && dc_step < id(${yambms_id}_var_auto_dcl_functions_count) + 1)
          {
            // Your C++ code
            // ...
            // ...
            // ...
            id(${yambms_id}_var_discharge_current_step)++;
          }
```

Currently, only the `Auto Temp CCL DCL` function uses this new `3-step loop`.

Here is the process for publishing the requested current : 

### Global vars

- `charge_current_step` = step counter
- `discharge_current_step` = step counter
- `auto_ccl_functions_count` = the number of `auto` functions that can reduce CCL and are part of the process
- `auto_dcl_functions_count` = same for DCL

### 3-step loop

- STEP 1: Calculation of the maximum charge/discharge current (YamBMS core lines 473 and 494)
  - `if (cc_step == 0 || cc_step == id(${yambms_id}_var_auto_ccl_functions_count) + 2)`
[...]

PR #212, opened 2026-02-07
Auto CCL and DCL to new loop

MrPabloUK:
Updated auto CCL and DCL functions, now full lambda integrated with the loop as described in #182.

Testing auto CCL with a full charge overnight.

Both files still use 'fast decrease, slow increase` logic, but the minimum increase step is now 1 amp.


MrPabloUK:
Auto CCL worked as expected, testing Auto DCL on a demo board today.

image

MrPabloUK:
Testing of DCL was successful, I consider this ready to review and merge.

PR #219, opened 2026-02-10
Auto CCL & DCL - Re-implement Soft Current Ramp

MrPabloUK:
After continued testing I've decided to re-implement the ramp up of current when the function is disabled.
Testing confirms that disabling the function during heavy charge or discharge can lead to large cell voltage changes, which in turn can trip BMS protection.


Sleeper85:
Can I merge it ?

PR #225, opened 2026-02-12
Add a document about charge current limits

haysdb:
Document the various charge current limits


Sleeper85:
This PR will be analyzed at a later stage based on our conversation on the topics of SEPLOS V3, Auto CCL Soft Landing and future tests.

PR #231, opened 2026-02-14
Add filters to Deye temperature sensors

GHswitt:
The Deye batteries are creating some noise on the temperature sensors as they only have full degree steps:

image

This commit adds a filter similar to the Basen one.

Draft: Currently testing.


GHswitt:
Will create one for more sensors

PR #243, opened 2026-02-22
Value checks

GHswitt:
Three commits with some minor fixes:
- Auto float: Cleanup code, no need for `double`
- Auto CCL/DCL: Check that `batt_capacity` is > 0 (startup/resets etc.)
- Core: Check variable to avoid division by zero
- Canbus: Replaced | with || in if clause

### DIFF
--- a/packages/yambms/yambms_core.yaml
+++ b/packages/yambms/yambms_core.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.02.19
-# Version : 1.7.0
+# Updated : 2026.02.22
+# Version : 1.7.1
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
 # YamBMS ( Yet another multi-BMS Merging Solution )
@@ -430,7 +430,7 @@ interval:
               // MEAN Battery SOC (Shunt/BMS)
               if (id(${yambms_id}_var_shunt_use) == true)
                 soc = round(id(${yambms_id}_var_shunt_total_soc) / id(${yambms_id}_var_to_combine_shunt_counter));
-              else
+              else if (id(${yambms_id}_var_total_battery_capacity) > 0)
                 soc = round(id(${yambms_id}_var_total_battery_capacity_remaining) / id(${yambms_id}_var_total_battery_capacity) * 100);
 
               // SOC manipulation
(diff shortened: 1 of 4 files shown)


MrPabloUK:
Thanks for the Auto Temp CCL / DCL change, agreed with the additional check.

packages/yambms/yambms_core.yaml
```diff
@@ -402,7 +402,7 @@ interval:
             if (id(${yambms_id}_var_to_combine_bms_counter) > 0)
             {
               // MEAN total_voltage (Shunt/BMS)
-              if (id(${yambms_id}_var_shunt_use) == true)
+              if (id(${yambms_id}_var_shunt_use) == true && id(${yambms_id}_var_to_combine_shunt_counter) > 0)
```
Sleeper85: This addition is unnecessary in my opinion; why add this condition ?

```cpp
            // +-----------------------------------------------+
            // | Uncombine not continuously                    |
            // +-----------------------------------------------+
            else if (id(shunt${shunt_id}_combined) == true)
            {
              // To combine shunt counter
              id(${yambms_id}_var_to_combine_shunt_counter) -= 1;

              // Shunt use (replace BMS infos)
              if (id(${yambms_id}_var_to_combine_shunt_counter) == 0) id(${yambms_id}_var_shunt_use) = false;

              // Uncombined
              id(shunt${shunt_id}_combined) = false;
            }

            // Increment the counter for next Shunt processing
            id(${yambms_id}_var_to_combine_shunt_number)++;
```

packages/yambms/yambms_core.yaml
```diff
@@ -428,9 +428,9 @@ interval:
                 battery_discharging_power = power * -1; // Must be positive energy
               
               // MEAN Battery SOC (Shunt/BMS)
-              if (id(${yambms_id}_var_shunt_use) == true)
+              if (id(${yambms_id}_var_shunt_use) == true && id(${yambms_id}_var_to_combine_shunt_counter) > 0)
```
Sleeper85: Idem, line 405.

Sleeper85 (changes requested):

packages/yambms/yambms_core.yaml
GHswitt: Only reason was: `${yambms_id}_var_to_combine_bms_counter` is checked above (line 402) before division. But `${yambms_id}_var_to_combine_shunt_counter` was not checked (at this place) before the division in line 406.

If it is not possible that `${yambms_id}_var_to_combine_shunt_counter` is zero at this point than this change can be removed.

packages/yambms/yambms_core.yaml
GHswitt: Same as the other one.

packages/yambms/yambms_core.yaml
GHswitt: I've removed this check

packages/yambms/yambms_core.yaml
GHswitt: I've removed this check.

PR #244, opened 2026-02-22
Auto CCL/DCL: Add division by zero/isnan checks

GHswitt:
@MrPabloUK Please check if this change is ok. I moved the `Exit if critical data is missing` check before the usage/division.

For the second check I'm not sure if `delta = -max_current;` is fine, the thought was if something is that wrong (cell_high_v - cell_float_v <= 0) then maybe stop charging.

### DIFF
--- a/packages/yambms/yambms_auto_ccl.yaml
+++ b/packages/yambms/yambms_auto_ccl.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.02.11
-# Version : 1.3.0
+# Updated : 2026.02.22
+# Version : 1.3.1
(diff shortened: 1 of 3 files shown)


MrPabloUK:
This change makes sense to me, thanks!
Will you be doing the same for AutoDCL?

GHswitt:
Yes, I can do that, will compare and see what applies. Lets use this PR, I set it to draft.

GHswitt:
@MrPabloUK Done for DCL, please check.

GHswitt:
Also looked at Auto CVL, added checks before variable calculations. Replaced double with float.

MrPabloUK:
Looks good to me, thanks again.

PR #248, opened 2026-02-24
Implement Auto CCL Soft Landing feature

haysdb:
# Auto CCL Soft Landing

Auto CCL Soft Landing is an optional, independent current-derating function that participates in the existing Auto CCL step pipeline.

---

## Purpose

Facilitate a controlled “soft landing” at bulk voltage by tapering requested charge current as pack voltage approaches bulk.

As pack voltage approaches bulk, voltage rises rapidly. Because the inverter and control logic do not respond instantaneously, the system may overshoot bulk voltage before current is reduced.

Soft Landing ensures the system reaches bulk under a controlled slope of gradually reduced current. When properly configured, it prevents bulk voltage overshoot and allows pack voltage to converge tightly to bulk while maintaining a small, predictable charge current near bulk (e.g., to give passive balancers time to operate).

This feature is optional and may not provide significant benefit in systems that already achieve stable convergence at bulk voltage. In systems where overshoot is managed by lowering bulk voltage, Soft Landing instead controls charge current directly so that the specified bulk voltage can be reached accurately without offset compensation.

---

## How It Works

When enabled, Auto CCL Soft Landing gradually reduces charge current as the battery approaches the configured bulk voltage.

- Charging proceeds normally until pack voltage reaches the knee voltage.
- Pack voltage must remain above the knee voltage for the configured dwell time before activation.
- Upon activation, the taper curve is anchored to the pack voltage at that moment to ensure a continuous transition without sudden current steps.
- Beyond the knee, charge current is progressively reduced as voltage approaches bulk.
- Near bulk voltage, a small, configurable “trickle” current is maintained.
- The function releases only when pack voltage falls below the knee minus hysteresis.

Charging completion (Cut-Off / EOC / Float) remains governed by YamBMS core logic.

---

## What It Does Not Do

- It does not use cell voltages or OVP thresholds.
- It does not implement termination logic.
- It does not replace Auto CCL or core charge-completion behavior.

---

## Optional Substitutions (defaults shown)
```yaml
auto_ccl_sl_knee_v:       '3.3875'  # knee/activation voltage (cell)
auto_ccl_sl_release_v:    '0.0125'  # hysteresis below knee to allow release (e.g., 0.2V/16)
auto_ccl_sl_activate_s:   '60'      # seconds voltage must remain >= knee to activate
auto_ccl_sl_knee_c_rate:  '0.08'    # minimum CCL (expressed as a c-rate) at the knee
auto_ccl_sl_bulk_c_rate:  '0.003'   # tail current for balancing
auto_ccl_sl_curve_exp:    '1.0'     # curve shape: 1=linear, 1.1..1.8 -> quicker decline, longer tail
auto_ccl_sl_shunt_spike:  '0.05'    # shunt voltage spike magnitude that will be damped
auto_ccl_sl_near_v_delta: '0.3'     # 'near bulk' delta voltage (pack) used in EMA blending
```
---

## Comparison with Existing Auto CCL

Auto CCL Soft Landing and the existing Auto CCL mechanism serve different purposes and operate on different inputs.

- **Auto CCL** limits charge current based on _cell_ voltage as cells approach the BMS over-voltage protection threshold. It is primarily a safety-oriented limiter and includes asymmetric response behavior to react quickly to rising cell voltage while avoiding oscillation.

- **Auto CCL Soft Landing** shapes charge current based on _battery_ voltage as the system approaches the configured bulk voltage. It provides a deterministic taper profile that allows maintaining a low, controlled charge current near bulk without relying on cell-level voltage thresholds.

Both mechanisms may be enabled simultaneously. When multiple charge-current limits apply, YamBMS uses the most restrictive result.

### DIFF
--- a/packages/yambms/yambms_auto_ccl_soft_landing.yaml
+++ b/packages/yambms/yambms_auto_ccl_soft_landing.yaml
@@ -26,20 +26,26 @@
 #    will result in Charging Instruction 'stop' before bulk voltage is reached.
 #  . Two-phase: curve (knee to bulk) ends at I_bulk (bulk_c_rate) at V_bulk; once V>=bulk,
 #    5s dwell then straight taper from I_bulk to 0 over ramp_to_zero_s.
+#  . After landed, CCL only rises again when V stays below (bulk - release_hyst_v) for
+#    release_dwell_s (debounce), so a single sample at the boundary does not release.
 #-------------------------------------------------------------------------------
 substitutions:
-  auto_ccl_sl_knee_v:         '3.4'    # knee/activation voltage (cell)
-  auto_ccl_sl_release_v:      '0.0125' # hysteresis below knee to allow release (e.g. 0.2V/16)
-  auto_ccl_sl_activate_s:     '10'     # seconds voltage must remain >= knee to activate
-  auto_ccl_sl_knee_c_rate:    '0.1'    # minimum CCL at the knee (expressed as a c-rate)
-  auto_ccl_sl_bulk_c_rate:    '0.01'   # c-rate at bulk: 6A / 614ah = .01c
-  auto_ccl_sl_curve_exp:      '1.0'    # curve shape: 1=linear, >1=quicker drop, longer tail
-  auto_ccl_sl_shunt_spike:    '0.05'   # shunt voltage spike magnitude that will be damped
-  auto_ccl_sl_near_v_delta:   '0.3'    # 'near bulk' delta voltage (pack) (used in ema)
-  # These substitutions define what happens at bulk
-  auto_ccl_sl_release_hyst_v: '0.02'   # pack V drop below bulk required before allowing CCL to rise again (after landed)
-  auto_ccl_sl_bulk_dwell_s:   '5'      # seconds V must remain >= bulk before starting ramp to 0 (5 readings at 1s)
-  auto_ccl_sl_ramp_to_zero_s: '30'     # seconds to ramp from bulk down to 0 after dwell (avoids voltage dip)
+  # Knee / activation (when we consider "near full" and start the taper)
+  auto_ccl_sl_knee_v:          '3.4'    # knee/activation voltage (cell)
+  auto_ccl_sl_release_v:       '0.0125' # hysteresis below knee to allow release (e.g. 0.2V/16)
+  auto_ccl_sl_activate_s:      '10'     # seconds voltage must remain >= knee to activate
+  auto_ccl_sl_knee_c_rate:     '0.1'    # minimum CCL at the knee (expressed as a c-rate)
+  # Curve (shape between knee and bulk)
+  auto_ccl_sl_curve_exp:       '1.0'    # curve shape: 1=linear, >1=quicker drop, longer tail
+  auto_ccl_sl_bulk_c_rate:     '0.0065' # c-rate at bulk (curve end, then hold then ramp)
+  # Bulk / landing (at and after bulk)
+  auto_ccl_sl_release_hyst_v:  '0.02'   # pack V drop below bulk required before allowing CCL to rise again (after landed)
+  auto_ccl_sl_release_dwell_s: '3'      # seconds V must remain below (bulk - hyst) before release (debounce)
+  auto_ccl_sl_bulk_dwell_s:    '5'      # seconds V must remain >= bulk before starting ramp to 0 (5 readings at 1s)
+  auto_ccl_sl_ramp_to_zero_s:  '30'     # seconds to ramp from bulk down to 0 after dwell (avoids voltage dip)
+  # Shunt / voltage smoothing (EMA and spike clamp)
+  auto_ccl_sl_shunt_spike:     '0.05'   # shunt voltage spike magnitude that will be damped
+  auto_ccl_sl_near_v_delta:    '0.3'    # 'near bulk' delta voltage (pack) (used in ema)
 
 esphome:
   on_boot:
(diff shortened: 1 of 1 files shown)


haysdb:
I'll make some 'tuning' changes before this goes live. I'm working on how to play nice with YamBMS cutoff logic. A long slow taper triggers an early transition to float so I'm experimenting with maintaining a higher charge current for longer along with an increase in the cutoff timer to give 'soft landing' a chance to finish charging before the transition to float. The good news is these are simply tuning parameters.

I'm also considering something I saw @MrPabloUK do in his LED code that I think is cool, to eliminate the 'debug'  sensors while keeping the data available to me in HA.

Sleeper85:
In discussion #213 I asked you to do a charging test following this procedure but you did not give any feedback, I don't know if you tried it ?

1. Set `Bulk` voltage to the desired voltage (the true target)
2. Do not set any time slots Deye TOU at 100%.
3. Set an `inverter_offset_v` of `0.5V` or more (only the Bulk v is considered for the target)
4. Fully charge your battery

What is the purpose of your Auto CCL Soft Landing function and why is it essential for you ?

I also have a Deye and I don't need the Auto CCL function to charge my battery properly.

haysdb:
When you say "only the bulk target is considered", considered by what? If CVL is 55.2 I get 55.4. If I add .2v inverter offset I get 55.6v. The inverter (more or less) obeys CVL. It does not know what my target voltage is. Are you referring to 'Auto CVL'?  It does't work for me because I have a battery with mismatched cells. The battery is balanced for SOC but at bulk voltage I have a large delta V due to IR differences in the cells. Auto CVL nerfs charging because it's trying to 'protect' a cell that doesn't need its protection.

My motivation is to charge my batteries with precision. My function charges to within .01v. This level of precision is unnecessary but the Seplos BMS does it so I want YamBMS to do it too.

After tomorrow I will perform your test. I'm trying to figure out an interaction with the cutoff logic. Charging to only 3.45v per cell makes it very easy for the cutoff voltage to fall below 3.37v. >10A and boom, cutoff voltage drops below 3.37v and restarts the cutoff timer. Took 20 minutes to reach EOC due to repeated restarts of the timer.

Sleeper85:
What I mean is that the Inverter_offset_v is not taken into account in the YamBMS charging logic; it's only used to reach the target voltage. I'm not asking you to use the current Auto CVL function, but simply to charge without any additional functions, correctly configuring your Deye without TOU 100% + Inverter_offset_v to reach the target bulk voltage.

haysdb:
> What I mean is that the Inverter_offset_v is not taken into account in the YamBMS charging logic; it's only used to reach the target voltage. I'm not asking you to use the current Auto CVL function, but simply to charge without any additional functions, correctly configuring your Deye without TOU 100% + Inverter_offset_v to reach the target bulk voltage.

OK, so just +.5 inverter offset and do not set TOU to 100%. The inverter should charge to .5v less than CVL to arrive at bulk. I'm still not understanding how these two scenarios are expected to be different:
1. SOC set to 100% in TOU. Bulk: 55.2v. Inverter offset: 0.
2. SOC not set to 100% in TOU. Bulk: 55.2v. Inverter offset: +.5v

I will perform the test after I negotiate the cutoff logic. I think that increasing bulk by .2v (which will increase the max cell voltage) and increasing the slope of the Auto CCL de-rate curve should allow 'soft landing' to pass through the area of 'turbulence' where cutoff repeatedly triggers and restarts. I also reduced the cutoff timer duration to leave a shorter window for the timer to be restarted. The longer the timer the better the chance of a current spike causing cutoff voltage to fall below 3.37v.

haysdb:
The cutoff logic does not consider bulk voltage at all, so if the inverter does not respect CVL, then there's nothing in core YamBMS to assure that the battery voltage stops rising at the bulk voltage, or even reaches the bulk voltage at all. All of the other Auto functions are 'protection'. Charging precision is what soft landing brings to the party.

Sleeper85:
Of course, the cutoff logic takes the bulk voltage into account. It's up to you to adjust your equipment correctly to achieve it.

haysdb:
> Of course, the cutoff logic takes the bulk voltage into account. It's up to you to adjust your equipment correctly to achieve it.

I stand corrected: bulk voltage is referenced once in the cutoff logic:
```
else if (max_cell_v < (cell_bulk_v - cutoff_resistance))
```

The cutoff logic is fragile. I am trying to navigate the logic but one instant of negative current resets the counter. A spot welder, for example. One pulse of current also resets the timer. With a high cell voltage of 3.46 it only takes 10 amps to drop cutoff voltage below 3.37. I'm seeing both of these. I have proposed a solution [here](https://github.com/Sleeper85/esphome-yambms/discussions/249)

It should be noted I was having this issue [long ago](https://diysolarforum.com/threads/yambms-jk-bms-can-with-new-cut-off-charging-logic-open-source.79325/page-202#post-1686918). Never figured out why it was happening because it's not easy to see _why_ charging status is toggling back and forth between bulk and cutoff:
Screenshot 2026-02-27 at 6 42 39 PM

haysdb:
> Of course, the cutoff logic takes the bulk voltage into account. It's up to you to adjust your equipment correctly to achieve it.

This is backwards. If battery voltage has reached bulk and current is low, and cutoff is failing, it's up to me to "adjust my equipment"?

haysdb:
A perfect run today:
- increased bulk voltage to 55.4v to keep cutoff voltage comfortably above 3.37v
- float: off (to make sure voltage reached bulk even without float)
- Equalizing: on (changed one of the balancers to make sure it came on)
- Changed the taper curve to keep current > 15A until very close to bulk to delay cutoff


Image 2-28-26 at 1 28 PM
Cutoff voltage stayed above 3.37 continuously. I think today it would have worked ok at 3.45v bulk.
Image 2-28-26 at 1 40 PM
No overshoot. Battery voltage topped out at bulk. .2v inverter offset just to make sure there's some 'push' at bulk.
Image 2-28-26 at 1 32 PM
Charge current and voltage wobbled around a bit. I might consider hysteresis near 0.
Image 2-28-26 at 1 32 PM (1)
Balancing of the high cell
Image 2-28-26 at 1 30 PM
I reduced the cutoff timer to 30 seconds to try to get through cutoff but I'll set it back to 60.
Image 2-28-26 at 1 36 PM

haysdb:
Still working to get consistent results. Some days it works perfectly, but other days it doesn't.

haysdb:
Getting this to work reliably requires changes to cutoff logic, which I think would not be considered, so I'm closing it.

Sleeper85:
Okay, feel free to give your opinion on the cutoff logic, I'm always open to improving where possible.

PR #261, opened 2026-03-14
Board file for LilyGo T-Connect Pro Lite (without screen)

davidcoulson:
Board file for LilyGo T-Connect Pro Lite (without screen)

### DIFF
--- a/board_ESP32-S3_LilyGo-T-Connect-Pro-Lite.yaml
+++ b/board_ESP32-S3_LilyGo-T-Connect-Pro-Lite.yaml
@@ -1,5 +1,5 @@
-# Updated : 2026.01.14
-# Version : 1.1.3
+# Updated : 2026.03.15
+# Version : 1.1.1
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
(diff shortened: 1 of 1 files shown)


Sleeper85:
Why not name this YAML `board_ESP32-S3_LilyGo-T-Connect-Pro-Ethernet.yaml` as it is the version with Ethernet instead of Display ?

davidcoulson:
> Why not name this YAML `board_ESP32-S3_LilyGo-T-Connect-Pro-Ethernet.yaml` as it is the version with Ethernet instead of Display ?

I think they both have Ethernet but Lite is what LilyGo call it. 

https://lilygo.cc/en-us/products/t-connect-pro

Scroll down and you can see.

Sleeper85:
Ah ok, I hadn't realized that it was actually a Lite version of the Pro board.

davidcoulson:
Is there a standard for boards with buttons - This has two buttons instead of the screen, but wasn't sure if there was a function that was typically mapped to the buttons?

Sleeper85:
> Is there a standard for boards with buttons - This has two buttons instead of the screen, but wasn't sure if there was a function that was typically mapped to the buttons?

No

PR #262, opened 2026-03-16
Add functions summary card

GHswitt:
Before I change all dashboards, here the idea: Summary for different (Auto) functions:

image

Only active/enabled functions are shown. Uses some colors if not active/paused/stopped e.g. for derating etc.

### DIFF
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
@@ -1,4 +1,4 @@
-# Updated : 2026.03.24
+# Updated : 2026.04.12
 # Version : 1.7.0
 # GitHub  : https://github.com/Sleeper85/esphome-yambms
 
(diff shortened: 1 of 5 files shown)


Sleeper85:
That's a good idea.

I've just finished working on the entity (sub-device) names and I've added an `Auto Functions` tab to all the dashboards for clarity. You can add your card there.

The switch `Force Bulk (top bal)` is renamed `Force Bulk`.

YamBMS becomes a device like any other; entity names simply begin with `${yambms_name}`.

GHswitt:
I've updated the Demo dashboard with the summary. If you think this is ok this way I can update the other dashboards.

Sleeper85:
Nice, it would look great on the front page.

Great work.

image

GHswitt:
I have it on my front page :-)

image

Sleeper85:
Since there's a lot of free space in the 3rd column, we could move entities from column 2 to column 3 and improve column 2 (add Balancer above Shunt by enlarging the area, as with the BMS).

* Rebulk Voltage
* ESPHome version
* YamBMS running version
* YamBMS latest version
* Loop time
* Uptime

@Haysdb suggested new sensors to understand what happens during the Cut-Off phase (cutoff V, cutoff A, timer).

What else ? 😆

image

GHswitt:
Updated the dashboards:

image

The disadvantage with the markdown table is that you can not click sensor values to get the history graph. So I would not move sensors for which you would like to open the history more often.

There is not yet a Balancer count sensor, also I did not look into the other sensors mentioned.

Sleeper85:
Thank you, I see there are still some points to improve.
It might be simpler if I take care of them myself.
I will then publish version **1.7.0**.

Sleeper85:
Done

PR #290, opened 2026-07-02
Add grace period to Auto EOC

GHswitt:
When the target time has passed for `Auto EOC`, the current limit is removed and the current jumps up to the max. charge limit, for example from 20A to 200A.

This is undesirable when only a small amount is missing. The change allows an one hours grace period. After the hour is up, it will return to the stopped state and there is no limit.

I'm currently testing it, therefore I will keep it as draft. Done testing. The feature is now quite improved.
Before the requested charge current was often a sawtooth since the calculation was done on integral steps of the capacity:
image

Now with light coloumb counting it is much better, as shown here. It was cloudy at the end and it adjusted accordingly, but still no up and down like before. Middle chart shows the calculated capacity compared to the integral (reported) capacity steps:
image

With the grace period now the battery will charge gently at the end even when the time is up.


Sleeper85:
Could you please rebase your branch so that it only contains the changes you made ?

Thanks

GHswitt:
Ok rebased again.

PR #297, opened 2026-07-13
Haysdb current taper

haysdb:
Add Auto CCL Current Taper package

Custom Auto CCL pipeline step that tapers requested charge current between a configurable knee voltage and bulk voltage, writing only a delta to var_auto_custom_ccl (same contract as other Auto CCL packages).

Behavior:

Activates when pack voltage reaches the knee; releases at rebulk voltage
Maps V_knee→V_bulk to I_knee→I_bulk via a tunable power curve (C-rates × capacity)
Forces target current to 0 A after cutoff (var_eoc)
Uses a short SMA of total_voltage so CCL does not chase raw pack-V noise
Also exposes enable switch, knee/bulk C-rate and curve-exp numbers, and
diagnostic sensors for SMA voltage, delta, and derived knee/bulk amps.

### DIFF
--- a/packages/yambms/yambms_auto_ccl_current_taper.yaml
+++ b/packages/yambms/yambms_auto_ccl_current_taper.yaml
@@ -54,8 +54,8 @@ number:
     name: "Auto CCL CT Knee Voltage"
     unit_of_measurement: V
     device_class: voltage
-    min_value: 54.2
     max_value: 56.0
+    min_value: 54.2
     step: 0.1
     mode: ${yambms_input_number_mode}
     restore_value: true
@@ -70,8 +70,8 @@ number:
     device_id: yambms_yambms${yambms_id}
     name: "Auto CCL CT Knee C-Rate"
     unit_of_measurement: "C"
-    min_value: 0.05
     max_value: 0.5
+    min_value: 0.05
     step: 0.005
     mode: ${yambms_input_number_mode}
     restore_value: true
@@ -86,8 +86,8 @@ number:
     device_id: yambms_yambms${yambms_id}
     name: "Auto CCL CT Bulk C-Rate"
     unit_of_measurement: "C"
-    min_value: 0.005
     max_value: 0.1
+    min_value: 0.0
     step: 0.005
     mode: ${yambms_input_number_mode}
     restore_value: true
@@ -102,8 +102,8 @@ number:
     id: yambms${yambms_id}_number_auto_ccl_ct_curve_exp
     device_id: yambms_yambms${yambms_id}
     name: "Auto CCL CT Curve Exp"
-    min_value: 0.5
     max_value: 2.0
+    min_value: 0.5
     step: 0.1
     mode: ${yambms_input_number_mode}
     restore_value: true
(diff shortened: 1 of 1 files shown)


Sleeper85:
Please update the 5 dashboards example and add your function to the `Auto functions` tab with a `README`.

If you plan to offer other charging profiles in the future, perhaps `Current taper` could be a profile selectable from a menu.

In that case, your function could be called `Auto CCL charging profiles`, this is just an idea.

image

haysdb:
It's a good idea. Since the knee voltage, the C-rates for the beginning and end, and the curve shape are all configurable, it already pretty flexible. If there is some advantage to a stepped profile like Pylontech does I could add that. All are just different ways to taper current so the name seems descriptive, but if someone can document how different BMSs work, I could add preset profiles for Pylontech, Seplos, JK, etc.

I'm uncertain as to how to 'deliver' this function. It occupies the Custom CCL 'slot' meaning it could be replaced by another custom CCL function. I'm handling it in my dashboard with a conditional so that the entire card is only displayed if the function is compiled into the firmware. Soft Landing also uses the custom CCL slot. There could be more than one compiled into the firmware but only one should be active at any one time since they all use the same slot. At present I just have soft landing commented out in my config file.

What do you think? All compiled in but left to the user to only activate one at a time? The regular Auto CCL can be active at the same time. It's only the custom CCLs where only one should be active at the same time unless the user manages which one is in control.

I mention this because I actually have THREE Auto CCL functions. One controls charging up to 98%. That's actually using the Auto EOC slot. One controls the final 2% (current taper). And then Auto CCL is there to handle 'protection' although I have that disabled.

haysdb:
> Please update the 5 dashboards example and add your function to the `Auto functions` tab with a `README`.

I don't know how to do that. I know about dashboards but not how to deal with one in GitHub.

I'm not using your dashboards. I started with yours but mine has drifted far away.

Sleeper85:
It's simple: in your YamBMS fork, you must first create a synchronized branch on `dev` dedicated to your PR, such as `PR_Current_Taper`.

You work on this branch to add/modify the YAML files that need updating. For ease of use, flash an ESP32 with `YamBMS_RP_DEMO.yaml (dev branch)` and add the `DEMO dashboards` to your HA to add a card dedicated to your `Current Taper` function.

Then, update all the example dashboards by adding the missing card (the few lines of YAML that make up your card).

Your card must include a `weblink` entity with a link to your function documented in `YamBMS_functions.md`.

Your PR will then be complete and ready to merge.

Thanks

haysdb:
Oh boy. OK. I don't have a spare ESP32 with CAN and RS485 plus I'm not where my HA machine is to do the flash. I created a new branch and created the new file but I think I'm stuck without being able to view the native dashboard. I can still maybe create it using the demo naming convention.

Sleeper85:
The DEMO version doesn't require CAN or RS485; it's code with fake BMS and shunt.
A simple ESP32 is sufficient.

You don't even need to use a DEMO ESP32 to update the dashboards but it's just easier to see the result and be sure that everything is OK.
Create a dedicated `card` for your function and then add the relevant YAML section to the example dashboards.

haysdb:
Add it to all of the dashboards? I put it as the last card.

Sleeper85:
In the `Auto functions` tab, in the red box, that will be perfect. Don't forget the link to the REDAME (weblink entity) explaining how your function works.

image

Sleeper85:
Update your branch before modifying the dashboards because I updated them myself this morning.

haysdb:
I created a new PR

PR #303, opened 2026-07-15
Current Taper

haysdb:
I'm sorry I'm wasting your time.

I deleted my repository, re-forked, created a branch, added one file, edited the document, and added a new entities card to each of the 5 dashboards.

If this still contains changes to other files I don't understand how.

If it is missing edits, I don't know how because I created a new fork.


Sleeper85:
Please rename `yambms_aut_ccl_current_taper.yaml` to `yambms_auto_ccl_current_taper.yaml`

Sleeper85:
Sorry, to answer your request I've just updated `yambms_core.yaml`.

You now need to perform a rebase:

- In GitHub Desktop: **Branch → Update from upstream/dev** (available if you selected "contribute to parent project" when cloning)
- Or: **Current Branch → choose your branch → Branch → Rebase current branch...** onto `upstream/dev`
- Then **Push origin** (Desktop will prompt for a force push if you rebased)

Sleeper85:
Unless absolutely necessary, I will no longer make any changes to the YAML files related to the Cut-Off charging logic.

Please complete your PR as required, and then I will merge it.

Sleeper85:
Thank you for replacing the **weblink name** `Auto CCL Current Taper docs` with `README` simply.

Sleeper85:
### Thank you for taking these comments into account.

1. **`Knee Voltage` limits are hardcoded for 16S LFP.** `min_value: 54.2`, `max_value: 56.0` and `initial_value: 54.4` make the entity unusable on 8S packs, Li-ion or LTO. YamBMS already has a pattern for this: compute the traits at boot from `cell_count × chemistry values`, like `bulk_voltage` does. See the `on_boot` block below.

2. **`knee_v > bulk_voltage` is currently possible** (knee max 56.0 vs a lower bulk setting). In that case `bulk_latched` fires at activation and the function jumps straight to the post-bulk floor. Not dangerous, but surprising. Suggestion: a simple runtime guard treating the function as a no-op when `V_knee >= V_bulk - 0.2f` (the boot traits can't cover it since `bulk_voltage` is adjustable at runtime). Side note: the 0/0 division in `p` that this case would produce is already neutralized by accident — `fminf/fmaxf` evacuate the NaN into the clamp.

3. **`Post Bulk Amps` vs the Cut-Off current deadband.** The number allows 0 A: below `0.005C` (1.4 A on a 280 Ah pack) the current falls into the Cut-Off deadband, the compensated criterion is no longer evaluated and the charge can only complete through the *fully charged at rest* signature. It works, but the doc should recommend keeping `post_bulk_a` above `0.005C × Ah` so the classic path stays available.

4. **After EOC with Float enabled**, `I_target = 0` holds as long as `var_eoc` is true and the session is active: Float cannot deliver current until the pack drops below `knee − 0.2 V`. Short transient in practice, and consistent with the "CVL not trusted" philosophy, but worth a line in the doc for Float users.

5. **Cosmetics:** two comments are out of sync with the code — "8-sample SMA" vs `window_size: 10`, and the `curve_exp` comment mentions "~0.3 … ~2.5" while the actual range is 0.5–2.0.

---

**Proposed `on_boot` for point 1** (replaces the existing block — the registration stays at `-100`, the traits run at `550`, i.e. right after the YamBMS chemistry block at priority `600` which sets the `var_cell_*` globals this code reads):

```yaml
esphome:
  on_boot:
    # Registration in the Auto CCL pipeline
    - priority: -100.0
      then:
        - lambda: |-
            id(yambms${yambms_id}_var_auto_ccl_functions_count)++;   // Registration

    # Knee Voltage limits based on cell count and battery chemistry
    # Priority 550 : runs right after the YamBMS chemistry block (600)
    # which sets the var_cell_* globals used below
    - priority: 550.0
      then:
        - lambda: |-
            const int cell_count = id(yambms${yambms_id}_cell_count).state;
            const float v_min = round(cell_count * id(yambms${yambms_id}_var_cell_nominal_v) * 10) / 10;
            const float v_max = round(cell_count * id(yambms${yambms_id}_var_cell_max_charge_v) * 10) / 10;

            id(yambms${yambms_id}_number_auto_ccl_ct_knee_v).traits.set_min_value(v_min);
            id(yambms${yambms_id}_number_auto_ccl_ct_knee_v).traits.set_max_value(v_max);

            // Initialize/clamp the value inside the pack window
            // (the static initial_value is 16S LFP), default = 95% of bulk knee zone
            const float knee_v = id(yambms${yambms_id}_number_auto_ccl_ct_knee_v).state;
            if (isnan(knee_v) || (knee_v < v_min) || (knee_v > v_max))
              id(yambms${yambms_id}_number_auto_ccl_ct_knee_v).publish_state(round(cell_count * id(yambms${yambms_id}_var_cell_min_charge_v) * 10) / 10);
```

haysdb:
> Unless absolutely necessary, I will no longer make any changes to the YAML files related to the Cut-Off charging logic.
> 
> Please complete your PR as required, and then I will merge it.

Thank you for making those changes. I have a custom yambms_core so I'm just trying to improve it for others.

haysdb:
> Sorry, to answer your request I've just updated `yambms_core.yaml`.
> 
> You now need to perform a rebase:
> 
> * In GitHub Desktop: **Branch → Update from upstream/dev** (available if you selected "contribute to parent project" when cloning)
> * Or: **Current Branch → choose your branch → Branch → Rebase current branch...** onto `upstream/dev`
> * Then **Push origin** (Desktop will prompt for a force push if you rebased)

I do not use GitHub Desktop. I didn't know such a thing existed. The web does not make it obvious how to do this but I'll work on it.

Sleeper85:
I have just merged PR #290 making changes to `YamBMS_functions.md`, please rebase your PR.

GHswitt:
After syncing dev you could create a new branch based on origin/dev. Then cherry-pick (and merge your commits when necessar). Squash similar commits together, then it is easier to follow. Then it should be good to go.

Sleeper85:
I'm away this weekend, I'll check it out soon.

haysdb:
I have made all of the requested changes. It will get tested within the next hour.

I made a new change to start the curve at the CCL in effect at the moment CT becomes active. If Auto EOC has reduced current, CT will pick up smoothly from there.

Sleeper85:
Okay, I'll let you test this for a few days, let me know when I can merge it.

haysdb:
> 3. **`Post Bulk Amps` vs the Cut-Off current deadband.** The number allows 0 A: below `0.005C` (1.4 A on a 280 Ah pack) the current falls into the Cut-Off deadband, the compensated criterion is no longer evaluated and the charge can only complete through the _fully charged at rest_ signature. It works, but the doc should recommend keeping `post_bulk_a` above `0.005C × Ah` so the classic path stays available.

This will prevent a balance current of 2A with a battery greater than 400ah. This might be OK. I need to figure out how to simulate balancing. Something I can turn it on just before bulk and have it on for 15 minutes.

haysdb:
I'm including a balance simulator because it might be useful to people trying to 'tune' the balance current for their inverter or for people with BMSs that don't report equalizing. They could force a 'balance hold' every day or write a Home Assistant automation to create a balance hold on a weekly or monthly schedule.

Current taper wasn't written with balancing in mind but it may be a useful 'off label' use for it. Since CCL gets slammed to a few amps at bulk, there is no 'absorption' so some taper will be useful to move absorption forward. How many amp hours are available can be controlled with the taper. Maybe tapering to 30A leaves enough amp hours 'missing' that the battery can 'charge' at 3A for two hours. Something like that.

Lucky people with inverters that will hold CVL with CCL=0 are lucky. This is for inverters with their own heuristics for deciding when charging is finished. By controlling SOC, CVL, and CCL, we should be able to get the best of any inverter out there.

Edit: moving the balance simulator to a separate file.
Edit 2: and a separate PR

Sleeper85:
The DEMO code already contains a few lines of code to simulate equalization. It's a good idea to add this function for BMSs that don't provide it, but I would do it differently by integrating it directly into the relevant BMSs. It would also make more sense to simulate equalization when max_cell_v reaches a given voltage, such as bulk_cell_v. Keep it as is for your tests. I'll think about a better integration later.

haysdb:
I moved this to a different fork because it's a little more complicated than I originally thought. I have active balancers but I can't make them come on whenever I want and they typically only run for a minute.

I think the maintain-a-low-current-for-balancing feature might be one of the primary use-cases for current taper, so I need to be able to test that.

I agree about the max cell. This could be the 'equalizing' flag that some BMSs don't have. I think I should basically duplicate what the Neey style balancers do, with a Start Balance V, Stop Balance V, Start Delta V, and Stop Delta V. For testing they can be set such that the 'balancer' always comes on.

haysdb:
You said no more changes to yambms_core.yaml for cutoff but I need an 'Auto Custom CVL'. I've been using inverter offset but the purpose of that sensor is to compensate for voltage drop.  My use for it is to maintain 'pressure' during current taper and I also need it to go to zero at bulk, otherwise the battery voltage keeps rising even with CCL=2.

Other Auto functions control only CCL or CVL but I need to control both. This can run in parallel with Auto CCL and Auto CVL - the more restrictive wins.

```yaml
  - id: yambms${yambms_id}_var_auto_custom_cvl
    type: float
    restore_value: no
    initial_value: '0.0'
```

```cpp
requested_charge_voltage = (id(yambms${yambms_id}_bulk_voltage).state
    + id(yambms${yambms_id}_charger_offset_v).state
    + id(yambms${yambms_id}_var_auto_cvl)
    + id(yambms${yambms_id}_var_auto_custom_cvl));
```

haysdb:
I need to keep testing this. It's working because it's fundamentally very simple, but there are a few permutations of what can happen at bulk that I'd like to test.

All of the info I've been able to dig up with various LLMs says Deye inverters trigger only on SOC=100 and ignore CCL=0 but mine also gets an EOC signal from CVL suddenly dropping from bulk + offset to bulk. If I leave SOC=98 and don't change CCL, the drop from CVL 55.7 to 55.2 is interpreted as EOC. Or maybe this is still the .5v offset thing at work. Whether Minimum SOC is set to 100 in Time of Use or not I still see the .5v offset.

Getting the battery charged smoothly to bulk is a solved problem. Controlling what happens at 100% for balancing is what I'm working on now.

Sleeper85:
The `.cursor` folder and the `.gitignore` file must be removed from this PR.

haysdb:
I'm testing something more with current taper. The current taper part works. It's the balance-hold part I'm seeing if I can get to work with my inverter. 'The internet' says this is impossible. I set an offset of .5v to get the inverter to charge to bulk, but if I leave offset of .5v the battery voltage keeps rising. But if I drop it to bulk the inverter interprets that as 'charging complete' so I'm seeing if lowering it to bulk + .1v will keep the inverter in charging mode. At that point I'm hoping there is a low current that will maintain the bulk voltage without too much rise or fall. Or maybe the inverter will switch to CV mode and maintain CVL.

I feel like this must be a solved problem but I haven't been able to find any evidence that it's been solved.

If I can just get the inverter to keep listening I could dynamically adjust current to maintain CVL, but first I need to keep the inverter from deciding charging is complete.

Sleeper85:
Your simple `Current Taper` function becomes increasingly complex by including CVL modification.

The `charger_offset_v` is intended to adjust for a true voltage offset between the charger and the batteries (voltage drop on the cables or calibration issues).

The `0.5V` Deye offset is a separate issue.

Please explain clearly why your function reducing CCL should modify CVL ?

Your function should not include code to address the problems with your Deye.

Sleeper85:
An `Auto CVL - Dynamic Charger Offset` function would be more elegant.

haysdb:
I agree. It's gotten too complicated.  It's complicated because I'm trying to do two different things and because my inverter does not obey CVL.

I'll strip it back down to the basics and modify only CCL. I'll assume a well-behaved inverter that obeys CVL.

haysdb:
Stripped back to only current taper plus a balance current at bulk.

Question: If EOC occurs prior to the bulk voltage being reached, what should CCL be set to? Balance current or 0? I'm setting it to the balance current but I don't have a strong opinion either way. Do you know what CCL needs to be set to for inverters that will 'float'? For my inverter it doesn't seen to matter what CCL gets set to after the inverter decides that charging is complete.

haysdb:
I am now working on a local copy of current taper because I need custom behavior for my inverter. I can make changes as requested but I am no longer actively making changes.

Sleeper85:
I don't have much time for YamBMS at the moment, I'll read and check the code ASAP.

Sleeper85:
I'm merging your PR, which will be slightly modified afterwards, for the new `yambms_core.yaml` that I'm currently preparing.

Sleeper85:
[Changelog about new Auto CCL Current Taper (1.1.1 → 1.1.2)](https://github.com/Sleeper85/esphome-yambms/blob/dev/documents/README/changelog/Changelog_Auto_CCL_Current_Taper_1.1.2.md)

Thank you for giving me your feedback.

PR #304, opened 2026-07-17
Add Fake Balancer package

haysdb:
## Summary
- Optional per-BMS package that infers Equalizing from that BMS's max/min cell voltage using Enerkey-style thresholds (starting / sleep / trigger / stop-diff), for BMS that balance but never report a balancing flag (e.g. Seplos).
- Cooperates with a live external balancer on the same `bms_id` (`Equalizing = real OR inferred OR force`); includes a stub package when no hardware balancer is present.
- Docs and HA dashboard cards; Simulate Equalizing is documented after Current Taper for merge order with that PR.

## Test plan
- [ ] Include `balancer_simulate_equalizing.yaml` for one `bms_id` (with Enerkey/modbus client or stub) and confirm compile
- [ ] Enable Sim Equalizing: assert Equalizing when max cell ≥ starting and delta ≥ trigger; clear when max < sleep or delta ≤ stop-diff
- [ ] Force Equalizing asserts while on, independent of cell thresholds
- [ ] With a live Enerkey on the same BMS, Equalizing stays true if either sim or Enerkey is balancing; releasing sim does not clear Enerkey's equalizing
- [ ] Confirm Cut-Off holds while Equalizing is true and EOC timer remains the charge-end ceiling

Made with [Cursor](https://cursor.com)

### DIFF
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
@@ -459,6 +459,8 @@ views:
             name: Enable Sim Equalizing
           - entity: switch.sim_equalizing_1_force_equalizing
             name: Force Equalizing
+          - entity: number.sim_equalizing_1_max_run_time
+            name: Max Run Time
           - entity: number.sim_equalizing_1_balance_starting_voltage
             name: Start Balance V
           - entity: number.sim_equalizing_1_balance_sleep_voltage
--- a/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
+++ b/HomeAssistant_Dashboards/YamBMS_HA_Dashboard_DEMO.yaml
@@ -451,6 +451,25 @@ views:
             name: README
             icon: mdi:book-open-variant
         title: YamBMS - Auto EOC
+      - type: entities
+        title: Simulate Equalizing
+        show_header_toggle: false
+        entities:
+          - entity: switch.yambms_demo_sim_equalizing
+            name: Sim Equalizing
+          - entity: switch.yambms_demo_force_equalizing
+            name: Force Equalizing
+          - entity: number.yambms_demo_sim_eq_offset
+            name: Sim Eq Offset
+          - entity: number.yambms_demo_sim_eq_hold
+            name: Sim Eq Hold
+          - entity: binary_sensor.yambms_demo_sim_eq_active
+            name: Sim Eq Active
+          - type: weblink
+            url: >-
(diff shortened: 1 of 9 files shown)


haysdb:
I will be testing this over the coming days. I created the PR now so that you could offer some guidance.

It looks like a regular balancer to YamBMS and can be used in parallel with a real balancer so that the settings for the real balancer doesn't need to be touched.

This is designed as a diagnostic tool for tuning the 'balancing current' parameter in 'current taper' but it could be used with BMSs that don't report equalizing to YamBMS to 'infer' that balancing is occurring.

haysdb:
Where would you like the dashboard?
Screenshot 2026-07-21 at 8 05 47 AM

The entity names follow the naming convention of the Neey / Heltec balancer package.

haysdb:
This is working. I'm using it and haven't made any changes to it in a few days.

Sleeper85:
I don't have much time for YamBMS at the moment, I'll read and check the code ASAP.

Sleeper85:
Your PR #303 already included changes to the **dashboards** and `yambms_functions.md`.
Since I merged PR #303, your PR #304 has conflicts.

Furthermore, I will be working on `yambms_core.yaml` this afternoon.
I therefore advise you to wait a while before rebasing this PR.

Sleeper85:
Hi @haysdb, and sorry for the long wait.

I'm closing this PR, but the idea is landing — you found the problem, and the model you proposed (start and stop on a cell spread, with hysteresis) is what makes the inference trustworthy. You're credited in the changelog.

What I couldn't merge was the implementation. It's built around your own setup, a NEEY that isn't running every day, and that coupling makes it heavier than what YamBMS needs for a BMS that simply never reports the flag: four number entities holding absolute LFP thresholds next to the bulk voltage the user has already configured, a stub package that only exists so the lambda's balancer${bms_id}_* references resolve, and a copy of the bms_base_balancer.yaml branch to add the OR.

The version I merged anchors on bulk_voltage / cell_count instead — the same derivation the core already uses for end of charge — so it follows the chemistry, the cell count and the configured bulk voltage on its own, with no absolute threshold to maintain. It extends the raw BMS flag rather than the combined one, so bms_base_balancer.yaml is untouched and a real balancer keeps priority when it's online. That covers your NEEY case: when it's off the inference takes over, when it's on it wins.

Enabled by default for Seplos V3, Seplos V1/V2, EG4 LL v2 and JK RS485 DISPLAY, and available for any other BMS. Full write-up, thresholds and how to override them per BMS:

https://github.com/Sleeper85/esphome-yambms/blob/dev/documents/README/changelog/Changelog_Fake_Equalizing.md

I dropped Force and Max Run Time — triggering a top-balance by hand when the hardware balancer is idle feels like a balancer feature rather than a BMS one. Happy to look at it as its own package if you still want it.

Please give it a try next to your NEEY and tell me how it behaves.

PR #318, opened 2026-08-18
Fix boot-transient sensors reporting fabricated 0/sentinel values instead of N/A

txubelaxu:
## Summary

While updating a 2x JK-BMS RS485 installation (Victron Cerbo GX, CAN-bus BMS LV profile) from an older config to a recent `main`, I found that several sensors in `yambms_core.yaml` publish a **fabricated `0`** (or a raw, un-reset sentinel value) during the boot window before any BMS has combined, instead of reporting "unknown" (`NAN`). This is a real regression vs. an older reference point in this project's history, and it has a real-world consequence: it produced spurious low-voltage / high-temperature alarms on the connected Victron Cerbo GX during every reboot, because the first CAN frames sent to the inverter mixed already-real fields with fields still holding their reset sentinel.

Everything below was verified on physical hardware (ESP32-S3, 2x JK-BMS via RS485 sniffer, MCP2515 CAN interface to a Victron Cerbo GX), across multiple real reboots, with debug-level (`ESP_LOGD`) traces and by decoding the actual CAN-bus bytes transmitted. Each case is documented separately with the exact evidence found.

All commits touch `packages/yambms/yambms_core.yaml`, except Case 6 which touches `packages/bms/bms_base_charging_cycles.yaml`.

---

## Case 1 — SOC / requested charge / discharge voltage report `0` instead of N/A (real regression)

**Commit:** "Report N/A instead of 0 when no BMS is combined (SOC, CVL, DVL)"

At an earlier point in this project's history (`recovery-point-2026-02-03` tag), the SOC combine logic already did the right thing:

```cpp
if (id(${yambms_id}_var_to_combine_bms_counter) == 0) soc = NAN;      // no BMS combined => report N/A
```

By the current `main`, this had regressed to:

```cpp
if (id(yambms${yambms_id}_var_to_combine_bms_counter) == 0) soc = 0;      // no BMS combined => sending 0%
```

Same pattern for `Requested Charge Voltage` / `Requested Discharge Voltage`:

```cpp
if (id(yambms${yambms_id}_bms_combined).state == 0) // no BMS combined
  requested_charge_voltage = 0;
```

A `0%` SOC or `0V` requested voltage is indistinguishable from a genuine reading by anything consuming these sensors (Home Assistant automations, dashboards, or the inverter itself over CAN) - it looks like real data, not "not available yet".

Confirmed this doesn't reach the inverter over CAN, because `yambms_canbus.yaml`'s 100ms send loop already gates on `var_initialized`, which itself requires `requested_charge_voltage`/`requested_discharge_voltage` `> 0` - `NAN > 0` and `0 > 0` both evaluate false, so this specific fix doesn't change CAN-bus gating behavior. It only fixes what's reported to Home Assistant / any automation reading these sensors directly during the no-BMS-combined window.

---

## Case 2 — `battery_charging_power` / `battery_discharging_power`: two related bugs

**Commits:** "Also report N/A for battery_charging_power/discharging_power" and "Fix battery_charging/discharging_power staying NAN forever"

Same class of bug as Case 1, not yet covered by that fix: these two sensors were declared `int battery_charging_power = 0;` / `int battery_discharging_power = 0;` and published unconditionally, so they also reported a fabricated `0 W` before any BMS combined.

First attempt: changed both to `float ... = NAN;`. This introduced a **new, separate bug**, found via a live reboot test: once real data existed, only *one* of the two directions ever got assigned (`if (power > 0) battery_charging_power = power;` / `if (power <= 0) battery_discharging_power = ...`), so the *inactive* direction stayed permanently `NAN` instead of correctly reporting `0` - e.g. `Battery Discharging Power` showed "unavailable" for the entire time the system was charging, even though "0 W discharging while charging" is a real, valid computed value, not missing data.

Confirmed via live log:
```
[S][sensor]: 'Battery Discharging Power' >> nan W   (repeated every ~1s, for the whole time the battery was charging)
```

Fix: both variables default to `NAN` (true "no BMS combined" case), but are explicitly reset to a real `0` as soon as we're inside the "at least one BMS combined" branch, before picking which direction actually applies:

```cpp
battery_charging_power = 0;
battery_discharging_power = 0;
if (power > 0)
  battery_charging_power = power;
else
  battery_discharging_power = power * -1;
```

---

## Case 3 — `Installed Battery Capacity` has its own unguarded lambda

**Commit:** "Report N/A for Installed Battery Capacity before any BMS combines"

This sensor doesn't go through the STEP2 combine block at all - it has its own independent `lambda:`/`update_interval:`, returning the raw accumulator directly:

```cpp
lambda: return id(yambms${yambms_id}_var_total_installed_battery_capacity);
```

The accumulator's `initial_value` is `0.0`, so this published a fabricated `0 Ah` until the first BMS ever combines (this accumulator is set-once per BMS via `combined_once`, so a simple sentinel check works here: a real installed capacity is never `<= 0`):

```cpp
lambda: |-
  if (id(yambms${yambms_id}_var_total_installed_battery_capacity) <= 0) return NAN;
  return id(yambms${yambms_id}_var_total_installed_battery_capacity);
```

---

## Case 4 — The entire STEP2 combine block was affected, not just SOC/CVL/DVL

**Commit:** "Comprehensively fix STEP2 combine block: N/A instead of fabricated 0"

After fixing SOC/CVL/DVL, I did a full live audit of the boot transient by triggering real reboots and capturing every sensor publish from t=0. On the very first combine pass after boot (before any BMS has reported), the following were **all** found publishing fabricated values:

```
'Total Voltage' >> 0.00 V
'Current' >> 0.0 A
'Power' >> 0 W
'Battery SoH' >> 0 %
'Charging Cycles' >> 0
'Battery Capacity' >> 0 Ah
'Capacity Remaining' >> 0 Ah
'Max Charge current' >> 0 A
'Max Discharge Current' >> 0 A
'Balance Trigger Voltage (MAX)' >> 0.000 V
```

Worse, several sensors are published **unconditionally**, outside the `if (var_to_combine_bms_counter > 0)` guard entirely, directly leaking their raw reset-sentinel values:

```
'Min Cell Voltage' >> 5.000 V      <- the MIN sentinel (impossibly low, never reset yet)
'Max Cell Voltage' >> 0.000 V      <- the MAX sentinel (impossibly low, never reset yet)
'Delta Cell Voltage' >> -5.000 V   <- max_cell_v - min_cell_v = 0.000 - 5.000, a physically impossible negative delta
```

Root cause of a related, separately-discovered bug: my *first* attempt at fixing SOC used `int soc = 0; ... soc = NAN;` - **assigning a `NAN` float to an `int` is undefined behavior in C++** (a `NaN` bit pattern isn't representable as an integer). This "worked" (produced `nan %` in the logs) on one build, then silently produced a literal `0 %` again on the very next build after an unrelated change elsewhere in the same lambda, with no logic change to the SOC code itself. This is why every local variable in this block is now declared `float`, defaulting to `NAN`, instead of relying on assigning `NAN` to an `int`.

Fix: every local variable (`total_voltage`, `current`, `power`, `soc`, `soh`, `charging_cycles`, `battery_capacity`, `capacity_remaining`, `max_charge_current`, `max_discharge_current`) changed to `float`, defaulting to `NAN`. A `has_combined_data` boolean now guards every sensor that used to publish unconditionally (`balance_trigger_voltage`, `min/max_cell_voltage`, `min/max_voltage_cell`, `cell_ovp`, `cell_uvp`) so none of them can leak a raw accumulator/sentinel value before real data exists.

---

## Case 5 — Min/Max temperature: sentinel leak even after `has_combined_data` is true

**Commit:** "Fix Min/Max temperature sentinel leak (100.0/-100.0 degC)"

Found via a second live reboot test, *after* the Case 4 fix was already deployed: `Min temperature`/`Max temperature` still leaked their reset sentinels (`100.0`/`-100.0`, from `initial_value: '100.0'` / `'-100.0'`), **even at a timestamp where `has_combined_data` was already true and other combined sensors (Total Voltage, Current, Power) were already showing real values**:

```
[S][sensor]: 'Total Voltage' >> 53.06 V     <- real
[S][sensor]: 'Current' >> 9.1 A             <- real
[S][sensor]: 'Power' >> 482 W               <- real
[S][sensor]: 'Min temperature' >> 100.0 °C  <- sentinel leak, not real
[S][sensor]: 'Max temperature' >> -100.0 °C <- sentinel leak, not real
```

Root cause: the JK-BMS reports temperature in a different RS485 frame type than cell voltage/current, so a BMS can be counted as "combined" (`var_to_combine_bms_counter > 0`) one or more cycles before its temperature reading has actually been folded into `var_min_temperature`/`var_max_temperature` by `bms_combine.yaml`. `has_combined_data` alone isn't a reliable proxy for "temperature has genuinely been aggregated this round".

Fix: check the sentinel value directly, in addition to `has_combined_data`:

```cpp
bool has_min_temperature = has_combined_data && (id(yambms${yambms_id}_var_min_temperature) < 100.0f);
bool has_max_temperature = has_combined_data && (id(yambms${yambms_id}_var_max_temperature) > -100.0f);
```

### Real-world impact of Case 5 specifically

This was the actual cause of spurious alarms on a connected Victron Cerbo GX ("Low battery voltage" x2, "High Temperature 29°C") during reboots, confirmed by exact timestamp correlation with reboot events across 3 separate incidents, and then confirmed resolved with 4 consecutive clean reboots after this fix was applied.

Confirmed with `ESP_LOGD` debug traces exactly at the moment `var_initialized` first becomes `true` (i.e. the moment CAN transmission to the inverter starts) - post-fix, every value is real and self-consistent at that exact instant:

```
[D][yambms_debug:1057]: ===> YamBMS initialized <=== : true
[D][yambms_debug:1059]: Battery Voltage : 53.232002 V
[D][yambms_debug:1060]: Battery Current : 26.542002 A
[D][yambms_debug:1061]: Battery Power : 1413.000000 W
[D][yambms_debug:1062]: Battery SOC : 56.000000 %
[D][yambms_debug:1063]: Battery SOH : 100.000000 %
[D][yambms_debug:1064]: Min temp. : 29.100000 °C
[D][yambms_debug:1065]: Max temp. : 29.900000 °C
```

And confirmed by decoding the actual CAN bytes transmitted immediately after (`0x355` = SOC/SOH, `0x356` = voltage), matching the computed values exactly:
- `0x355`: `38 00 64 00 ...` → SOC = `0x0038` = 56, SOH = `0x0064` = 100 - matches `Battery SOC: 56%` / `Battery SOH: 100%`
- `0x356`: `cb 14 ...` → voltage = `0x14cb` = 5323 (×0.01V) = 53.23V - matches `Battery Voltage: 53.232002 V`

---

## Testing

- Physical hardware: ESP32-S3, 2x JK-BMS (JK_PB2A16S20P) via RS485 sniffer, MCP2515 CAN interface, Victron Cerbo GX (CAN-bus BMS LV profile, 500 kbit/s)
- Multiple real reboots per fix iteration, captured end-to-end via `esphome logs` from t=0 (button-triggered restart and OTA-triggered restart both used)
- OTA checksum verified before/after every upload
- Verified compile-clean at every step
- Verified via decoded CAN-bus frame bytes matching computed sensor values exactly (Case 5)
- Confirmed zero Victron Cerbo GX alarms across 4 consecutive reboots after the final fix, vs. 3/3 reboots with alarms before it
- Case 6: verified separately via a real OTA reboot on the same hardware, confirmed via the Home Assistant history REST API (not just the chart UI, to rule out any rendering artifact)

---

## Case 6 — `charging_cycles` / `total_charging_cycle_capacity`: explicit NaN→0 coercion in a per-BMS template sensor

**Commit:** "Propagate NaN instead of fabricating 0 in charging-cycles sensors"

Found separately from the rest of this PR, while investigating a real "0" appearing in a Home Assistant history graph for `sensor.jk_bms_1_charging_cycles` across a device reboot. This one isn't in `yambms_core.yaml` - it's a per-BMS `platform: template` sensor in `packages/bms/bms_base_charging_cycles.yaml`:

```cpp
lambda: |-
  float cycles_raw = id(bms${bms_id}_charging_cycles_raw).state;
  if (isnan(cycles_raw)) cycles_raw = 0;
  return cycles_raw + id(bms${bms_id}_charging_cycles_offset).state;
```

`cycles_raw` is `NAN` for the ~20s window after a reboot before the underlying RS485/BLE BMS component has published a real reading - this line explicitly coerced that to `0` before publishing, same bug class as the rest of this PR, just at the per-BMS sensor level instead of the multi-BMS combine level. Confirmed live via the Home Assistant history REST API (bypassing any chart rendering) on a real OTA reboot:

```
sensor.jk_bms_1_charging_cycles: 199.0 -> unavailable -> 0.0 -> 199.0   (before fix)
sensor.jk_bms_1_charging_cycles: 199.0 -> unavailable -> unknown -> 199.0   (after fix)
```

The neighboring `total charging cycle capacity` sensor in the same file has the identical pattern and was fixed the same way. Fix: drop the `if (isnan(...)) x = 0;` line in both and let NaN propagate through the addition (`NaN + x == NaN`), consistent with every other fix in this PR.


Sleeper85:
Hello,

I'll analyze all of this. However, your pull requests must be to the development branch, never to main.

[Read how to make a pull request to development.](https://github.com/Sleeper85/esphome-yambms/blob/dev/documents/README/dev/How_to_perform_a_PR_to_the_dev_branch.md)

Sleeper85:
I understand your request. I will work on it over the next few days. I will get back to you regarding the tests with Victron.

Sleeper85:
Please do not create a new PR to `dev` for this; I will adapt the code from `yambms_core.yaml` myself. We can continue discussing it here.

Sleeper85:
I left you a comment last night to avoid creating another pull request for this. I've already modified `yambms_core.yaml` on my end. You'll be able to test it soon.

Sleeper85:
I have published a new version of `yambms_core.yaml` with values ​​set to `NAN` when no BMS is combined.
See this  [this commit](https://github.com/Sleeper85/esphome-yambms/commit/dee203e42c31adba743ce6daf4e1cf8befee5aaf).
Thank you for testing this version and giving me your feedback.

Regarding the Victron **0x379 - Installed capacity** value, here is what the Victron protocol says :

>This is different from the similar parameter in 0x35F. In 0x35F in bytes 4 & 5 the usable capacity in Ah
is given as a UN16. At 48V this translates to maximum possible capacity of just over 3MWh, which
should be sufficient for the foreseeable future. It is recommended that batteries send the usable
online capacity: if one or more modules have disconnected, send a lower total usable capacity.

> The purpose of 0x379 is to show the installed capacity, so that the service engineer can see how
much capacity is offline.

The total installed capacity should therefore only decrease if you reduce the number of batteries in a fixed manner.

txubelaxu:
Tested `dee203e4` (your `yambms_core.yaml` on `dev`, current HEAD `f7170d4`) on my real hardware: ESP32-S3 Freenove, 2x JK-BMS via RS485 sniffer (`txubelaxu/esphome-jk-bms` fork), CAN to inverter.

Result: works as expected. During the boot/reconnect transient, before both BMS finish combining, `Battery SoC`, `Battery Capacity Remaining`, `Battery Discharging Power` and `Auto EOC Current` all publish as `nan` — no fabricated `0` anywhere — and transition cleanly to real values once `BMS combined` reaches 2. Clean boot, CAN bus responding normally, no crashes.

I only had to swap the sniffer's `external_components` source to my own fork (already the case in my production config, unrelated to your fix) to get it building against `dev`; no other changes were needed on my end.

Sleeper85:
Perfect

txubelaxu:
Correction / addendum to my test report above — I need to walk back part of it.

My "no fabricated `0` anywhere" claim was too broad. It was based on an `esphome logs` capture that only attached ~10s after the device rebooted, so it missed the very start of the boot transient. Home Assistant, being persistently connected, did record it: `sensor.jk_bms_1_charging_cycles` dips to `0` for a few seconds during the test window and then recovers to its real value of `201`.

**To be clear: this is not a regression from `dee203e4`, and not a problem with your fix.** `dee203e4` only touches `yambms_core.yaml`, and everything I reported about it stands — `Battery SoC`, `Battery Capacity Remaining`, `Battery Discharging Power` and `Auto EOC Current` all correctly publish `nan` through the transient, with no fabricated `0`.

The `0` comes from a different file that `dee203e4` doesn't cover, and which is still present on `dev` today — `packages/bms/bms_base_charging_cycles.yaml`:

```yaml
lambda: |-
  float cycles_raw = id(bms${bms_id}_charging_cycles_raw).state;
  if (isnan(cycles_raw)) cycles_raw = 0;      # <-- fabricates the 0
  return cycles_raw + id(bms${bms_id}_charging_cycles_offset).state;
```

and the same pattern a few lines below in `bms${bms_id}_cycle_capacity` for `capacity_raw`. `*_raw` is NaN until the first trusted BMS frame arrives after a reboot, so both sensors publish a literal `0` instead of going unavailable. With `state_class: total` on a counter, that `0` also lands in HA's long-term statistics.

The fix is just to delete both coercion lines and let NaN propagate (`NaN + x == NaN`), which makes the sensors report unavailable exactly like the ones you fixed in `yambms_core.yaml`. This was part of the diff in this PR, which never landed since you implemented the core fix directly on `dev` — so if you want it, it's a two-line change and I'm happy to open a small dedicated PR against `dev` for just this file. Just let me know.

Sorry for the incomplete first report — I should have checked the HA history rather than only the log capture.

Sleeper85:
Noted, I will check that part.

Sleeper85:
Done with this [commit](https://github.com/Sleeper85/esphome-yambms/commit/ff4991862809c063059fb49d74d0e784ca8f26c2).

